Issue #411, opened 2026-05-18
Cache analyzer: chunk authoring advisories (budget overflow + ordering)


spinje:
## Context

Follow-up to the multi-breakpoint prompt caching change. The runtime now places up to 4 `cache_control` breakpoints per request on Anthropic using first-N-individual + terminal-merge placement (`compute_marker_chunk_indices` in `src/pflow/core/cache_render.py`). Two analyzer advisories are deferred from v1 — both about authoring decisions the runtime can't fix.

These are bundled in one issue because:
1. They share the same data source (workflow IR + all `prompt_cache:` subsets across workflow + sub-workflows)
2. They are both INFO-tier advisories produced by a single analyzer pass
3. They both address "the runtime can't reorder or expand chunks; only the author can"

### Problem A — silent budget loss

Today, an author declaring 6 `## Cache` chunks on Anthropic (budget=4) gets:
- chunks 0, 1, 2 → individual cache breakpoints
- chunks 3, 4, 5 → merged into the terminal cache

The author may have expected 6 independent caches. There is no analyzer signal that chunks 3 and 4 lost their independence.

### Problem B — suboptimal chunk ordering

The benefit of multi-marker placement depends on chunks being declared **stable-to-volatile** (rare-change content first, frequent-change content last). When a workflow declares chunks volatile-first:
- Multi-marker placement still fires (no error, no warning)
- But every per-chunk cache invalidates on every call
- The terminal cache still works, so caching technically functions
- The author has no signal that they're missing the per-chunk reuse benefit

The convention is documented in `pflow guide caching` ("Order chunks stable-to-volatile") but documentation alone won't help authors who don't realize they violated it.

## Proposed solution

Two new catalog entries, both INFO severity, emitted from a single new analyzer pass.

### `cache.too-many-chunks-for-provider` (Phase 1, simple)

**Detection** (purely static, from IR):

For each LLM node, if `len(prompt_cache_subset) > get_breakpoint_budget(provider)` (after subtracting 1 if `prewarm: true`):

```
[cache.too-many-chunks-for-provider]
Node 'summarize' declared N=6 chunks but provider 'anthropic' supports M=4
independent cache breakpoints per request. Chunks 0..2 each get an individual
cache; chunks 3..5 are merged into the terminal cache.

Fix options:
  1. Consolidate to ≤4 chunks (combine related content).
  2. Move the least-reusable chunks (3..5 in current declaration order) to
     the END so the merge happens on volatile content, not stable content.
[...]

Issue #410, opened 2026-05-18
Per-marker cache hit/miss attribution in pflow report (multi-breakpoint telemetry)


spinje:
## Context

Follow-up to the multi-breakpoint prompt caching change. Anthropic now receives up to 4 `cache_control` breakpoints per request, each creating an independent cumulative-prefix cache. On reads, Anthropic returns the **longest matching prefix** via a 20-block lookback window.

The provider's response includes ONE aggregate pair per call: `cache_creation_input_tokens` + `cache_read_input_tokens`. There is no per-marker decomposition from the wire. But because the longest matching prefix is what fires, we can **derive per-marker hit/miss** from `cache_read_input_tokens` + the known cumulative token count at each marker position.

## Problem

Today's trace and `pflow report` surface:

```
cache_creation_input_tokens: 1,204
cache_read_input_tokens: 3,847
```

An agent reading this for a 4-marker workflow has no way to tell:
- Whether all 4 markers hit, or just the first 2
- Whether the bottom marker is wasted (always misses)
- Whether a chunk near the top has high enough churn that its individual marker doesn't pay for itself

The per-marker attribution is **derivable from existing data plus marker positions** — we just don't compute or surface it.

### Phase 1 — per-call attribution in `pflow report`

Add per-marker hit/miss to the per-node report page (one row per marker). Example output:

```
LLM call — claude-sonnet-4-5
  Markers placed: 4 (chunks: system, knowledge, context, dynamic)
  Longest prefix matched: chunk 2 (knowledge)

  marker  chunk        cumulative_tok  hit?  delta_tok
  ------  ----------   --------------  ----  ---------
  0       system        2,341          yes    2,341
  1       knowledge     3,847          yes    1,506
  2       context       4,327          no       480
  3       dynamic       13,527         no     9,200

  Cache read total: 3,847 tok  (matches marker 1 cumulative)
  Cache write total: 9,680 tok  (chunks 2-3 newly cached)
```

### Phase 2 — aggregate marker effectiveness in `pflow analyze-cache --from-trace`

Over many runs, surface per-marker hit rates:

```
Marker effectiveness (across N=42 calls):
  marker 0 (system, ~2,341 tok)     hit 100% (42/42)
  marker 1 (knowledge, +1,506 tok)  hit  87% (37/42)
  marker 2 (context, +480 tok)      hit  12% (5/42)
  marker 3 (dynamic, +9,200 tok)    hit   0% (0/42)

[...]

Issue #409, opened 2026-05-18
Cache analyzer: warn when Anthropic-via-proxy models silently miss multi-breakpoint caching


spinje:
## Context

Follow-up to the multi-breakpoint prompt caching change (Task 159 DD#11 follow-up). The runtime now places up to 4 `cache_control` breakpoints per request on Anthropic, but only one on all other providers.

Provider detection lives in `src/pflow/core/llm_providers.py::detect_provider` and returns one of `"anthropic"`, `"openai"`, `"gemini"`, or `None`. The "or `None`" case is load-bearing here: models routed through proxies — `openrouter/anthropic/claude-sonnet-4-5`, `bedrock/anthropic.claude-sonnet-*`, `vertex_ai/claude-sonnet-*`, etc. — explicitly do NOT classify as `"anthropic"` because the prefix doesn't match. `get_breakpoint_budget(None)` returns 1, so the runtime silently downgrades to single-terminal-marker placement.

## Problem

A user routing Anthropic through a proxy sees:
- Identical workflow definitions
- Identical chunk declarations
- Identical traces (just the proxy prefix in the model name)
- But **75% of the multi-marker benefit silently lost**

`v1` ships a section in `pflow guide caching` documenting this ("Anthropic via proxies: per-chunk caching only fires with the `anthropic/` prefix"). That helps users who read the guide; it does nothing for users iterating on a workflow who don't realize the feature isn't engaging.

## Proposed solution

Add a new advisory `cache.routed-provider-degraded` emitted by the cache analyzer (NOT the runtime — runtime spam per call is hostile).

### Detection

In `src/pflow/core/cache_analysis/analyze.py`, during per-call row construction. Fire the advisory when ALL of:

- `n_rendered_chunks > 1` (workflow declared multiple chunks → multi-marker WOULD help)
- `get_breakpoint_budget(provider_name) == 1` (currently not getting multi-marker)
- The model identifier looks Anthropic-shaped despite not matching the canonical prefix. Match heuristic candidates:
  - Contains `/anthropic/` (OpenRouter style)
  - Contains `anthropic.claude` (Bedrock style)
  - Contains `claude-` after another provider prefix (Vertex `vertex_ai/claude-*`, future proxies)

The match heuristic should be conservative (false positives are noisy) — start with the three explicit patterns above; treat unknown shapes as not-matching.

### Severity

INFO. The user's workflow still runs; caching still works (just as single terminal prefix). This is an opportunity advisory, not a runtime failure.

[...]

PR #412, opened 2026-05-18
feat: multi-breakpoint prompt caching for Anthropic (Task 159 follow-up)

spinje:
## Summary

Place up to 4 `cache_control` breakpoints per Anthropic request — one per declared chunk — so each cumulative prefix becomes independently cacheable. Other providers unchanged. The hash side is intentionally untouched: marker placement is a wire-format concern, so memo cache identity and DD#19 byte-identity are preserved automatically, golden baseline does not drift.

Before this change: workflows declaring N `## Cache` chunks got ONE cache_control marker on the last block (v1 single-breakpoint strategy, DD#11).

After this change (Anthropic): first `(budget - 1)` chunks each get their own marker; remaining chunks merge into the terminal marker. Budget shrinks by 1 when `prewarm: true` consumes a slot. Other providers (OpenAI, Gemini, unknown) get a terminal marker only — byte-identical to today.

## Task

159

## Changes

- `core/llm_capabilities.py` — `get_breakpoint_budget(provider_name)` with `ANTHROPIC_BREAKPOINT_BUDGET=4` / `CONSERVATIVE_BREAKPOINT_BUDGET=1`
- `core/cache_render.py` — `compute_marker_chunk_indices()` pure deterministic placement; raises `ValueError` on empty rendered list (caller contract)
- `nodes/llm/llm.py` — `_build_system_blocks` placement loop replaces single `blocks[-1]` write; `_strip_below_min_cache_markers` suppresses the `cache.below-min-rendered` warning when any marker in a channel survives (avoids the new noise regression that multi-marker would otherwise introduce on multi-chunk workflows with small early chunks)
- `core/cache_analysis/analyze.py` + `cache_analysis/CLAUDE.md` — `_GroupedConsumerProjection.per_call_prefix_tokens` docstring updated; logic unchanged
- `core/llm_client.py` — adapter docstring updated to reflect plural-marker reality
- `guide/features/prompt-caching.md` — two new sections: "Order chunks stable-to-volatile" and "Anthropic via proxies: per-chunk caching only fires with the `anthropic/` prefix"
- `examples/core/prompt-caching-multi-chunk.pflow.md` — new 3-chunk fixture
- Stale single-marker docstrings updated at 5 sites in lockstep

## Explanation

Anthropic supports up to 4 `cache_control` breakpoints per request, and crucially the pricing model is **delta-based on writes** — adding more breakpoints does not cost more. On reads, the longest matching prefix wins via a 20-block lookback. So multi-marker placement is strictly an improvement: writes are free vs. single-marker, reads gain independent per-chunk reuse.

The architectural insight that made this small (~50 LOC production): the hash side at `runtime/engine/plan_node.py::_render_cache_for_hash` produces `{"name", "prose", "value"}` dicts with NO `cache_control` keys. Markers are a pure wire-format concern. So multi-marker placement adds keys to the prep-side blocks but does not touch the hash bytes — golden hashes, memo cache entries, and the DD#19 byte-identity contract are preserved automatically with no migration story.

The plan was reviewed by 4 specialized review agents (plan structural, impact completeness, feature interactions, silent failures) before implementation. Two real plan-level errors were caught during implementation:
- `assert n_rendered_chunks >= 1` would be stripped under `-O` (ruff S101) — switched to `raise ValueError(...)` for loud failure on caller-side bugs
- A planned test for "declared survives + prewarm stripped" turned out to be structurally unreachable because the strip walker accumulates cumulative tokens across both channels in order — reframed to verify the same suppression rule on prewarm-only configurations

Stats: 15 files changed, 1,178 insertions, 25 deletions. ~50 LOC production code, ~280 LOC tests (39 new tests).

## Created Docs

- `scratchpads/multi-breakpoint-prompt-caching/progress-log.md` — full implementation log, deviations from plan, test gates, risk register

User-facing documentation updates:
- `pflow guide caching` — two new sections explaining the stable-to-volatile chunk ordering convention and the proxy-prefix limitation

## Testing

New tests (39, all pass):

- `TestComputeMarkerChunkIndices` (15) — verifies the placement algorithm across every provider/budget/prewarm permutation including the contract-enforcement `ValueError` on empty input
- `TestGetBreakpointBudget` (5) — Anthropic=4, others=1, None=1, unknown=1
- `TestMultiBreakpointPlacement` (10) — end-to-end through `LLMNode.run()`: 3-chunk individual, 7-chunk first-3-plus-terminal, prewarm budget reduction, OpenAI/Gemini terminal-only, user_system offset, ABSENT-chunk count, fresh-dict-per-block, prewarm-bailed-but-still-reserves, user_system contributes to strip threshold
- `TestStripBelowMinSuppression` (4) — verifies the warning fires when ALL markers in a channel were stripped (true failure) and is suppressed when any marker survives
- `test_cached_system_renders_all_multi_breakpoint_markers` — `pflow report` smoke test for multi-marker output

Locked existing contracts (must continue passing unchanged):

- `test_golden_baseline_hashes_match` — golden hashes do not drift
- `test_hash_render_and_prep_render_byte_equivalent_through_namespaced_store` + `_with_absent_chunks` — text byte-equivalence preserved (markers are not text)
- `test_rendered_strip_does_not_mutate_text_bytes` — strip path was already multi-marker aware
- 377 cache_analysis regression tests

Manual verification:

- `make check` clean (ruff/ruff-format/mypy/deptry)
- `make test` — 6,998 passed, 1 skipped
- `pflow examples/core/prompt-caching-multi-chunk.pflow.md --dry-run article=... session_id=...` validates clean (5 nodes, 2 LLM with prompt_cache)
- `pflow analyze-cache` on the new example workflow produces a clean cache report

Live API verification (Anthropic + OpenAI, ~$0.10 total):

- **4-chunk Anthropic, fresh call**: 4 `cache_control` markers placed on the wire, 7,051 tokens written to provider cache
- **4-chunk Anthropic, second call**: all 7,051 tokens read from cache. $0.0266 → $0.0023 (~12× cheaper) end-to-end
- **Partial invalidation**: change chunk[2] between calls → `cache_read=3537` (chunks 0+1 still hit), `cache_write=3565` (chunks 2+3 fresh). Per-chunk independent caching confirmed against real provider
- **5-chunk over-budget Anthropic**: build placed 4 markers at `[0,1,2,4]`; strip removed block 0 (sub-threshold); 3 markers survived; suppression rule fired (no misleading warning)
- **`system:` param + 3 chunks**: `chunk_block_offset = 1` correctly applied; user_system block carries no marker; chunk markers placed on blocks `[1,2,3]`
- **OpenAI fallback**: pflow places terminal marker (verified via in-process instrumentation); LiteLLM strips before send (correct wire behavior for OpenAI)
- **`--dry-run`, `analyze-cache --from-trace`, non-contiguous subsets, empty/all-ABSENT chunks**: all pass

Reproducers and full trace data in `scratchpads/multi-breakpoint-prompt-caching/verify/`.

## Deferred follow-ups (filed as separate issues)

This PR ships the minimum that delivers the win. Five follow-ups are tracked separately:

**Scope-deferred (intentional choices to keep v1 simple):**

- #409 — `cache.routed-provider-degraded` advisory for Anthropic-via-proxy silent degradation
- #410 — per-marker hit/miss attribution in `pflow report` (multi-breakpoint telemetry)
- #411 — chunk authoring advisories (budget overflow + ordering recommendations)

**Surfaced during verification (pre-existing behaviors that multi-marker makes more visible):**

- #413 — LiteLLM mutates pflow's `system_blocks` list in-place, so the trace shows post-mutation state for non-Anthropic providers. Trace says "0 markers" even when pflow correctly placed one. Pre-existing v1 behavior; multi-marker doesn't trigger it because Anthropic preserves markers.
- #414 — `_count_text_tokens` underestimates Claude tokens by ~33% (LiteLLM's tokenizer falls back to `cl100k_base`). Pre-existing v1 bias documented as "intentional for Gemini safety" but causes Anthropic-only false-strip. Multi-marker amplifies the impact because earlier markers sit at lower cumulative tokens, more likely to land near the threshold.

Neither #413 nor #414 blocks the v1 ship — pflow's placement is correct and the wire format is correct for the configurations real users will hit. They're follow-ups to surface and resolve real but bounded edge cases.

Smart automatic placement (cross-node-aware marker positions) is intentionally NOT yet filed — it's contingent on #410's telemetry showing first-N-individual leaves wins on the table.


spinje:
# Multi-Breakpoint Prompt Caching for Anthropic — Implementation Progress Log

**Branch**: `main` (working tree)
**Date**: 2026-05-18
**Plan source**: User-supplied implementation plan (Task 159 follow-up, see DD#11 / DD#19)

## What this delivers

Workflows with multiple `## Cache` chunks that change at different
frequencies (system prompt → knowledge base → session context → per-call)
now get **automatic, per-chunk cache reuse on Anthropic** — without any
new workflow syntax and without affecting non-Anthropic providers.

Before this change, despite a workflow declaring N chunks, the runtime
emitted **one** `cache_control` breakpoint per LLM call (on the last
chunk). That was the deferred v1 single-breakpoint strategy.

After this change:
- **Anthropic** (budget=4): first `(budget - 1)` chunks each get their own
  marker; remaining chunks merge into the terminal marker.
- **All other providers** (budget=1): terminal marker only, byte-identical
  to today's behavior.

## High-level result

- **Production code change**: ~50 LOC across 3 files.
- **Test code change**: ~280 LOC across 4 test files.
- **Zero hash drift**: golden baseline tests pass unchanged.
- **Zero memo cache invalidation**: hash side intentionally untouched.
- **Final gates**:
  - `make check`: ruff/ruff-format/mypy/deptry all pass.
  - `make test`: 6998 passed, 1 skipped.
  - Targeted: 175/175 cache-render+capability+hash+rendering+strip tests pass.
  - Cache analysis regression: 377/377 pass.

## Files changed

### Production code

| File | Change |
|---|---|
| `src/pflow/core/llm_capabilities.py` | Added `ANTHROPIC_BREAKPOINT_BUDGET=4`, `CONSERVATIVE_BREAKPOINT_BUDGET=1`, `get_breakpoint_budget()`. |
| `src/pflow/core/cache_render.py` | Added top-level import of `get_breakpoint_budget`; added `compute_marker_chunk_indices()` pure function. Updated stale docstring on `_build_cache_control_marker`. |
| `src/pflow/nodes/llm/llm.py` | Imported `compute_marker_chunk_indices`. Replaced single-marker placement at `_build_system_blocks` with a loop over computed indices. Updated docstring. Extended `_strip_below_min_cache_markers` with per-channel warning suppression when any marker survives. |
| `src/pflow/core/llm_client.py` | Adapter docstring updated. |
| `src/pflow/core/cache_analysis/analyze.py` | `_GroupedConsumerProjection` docstring updated (per-call denominator stays correct under multi-marker). |
| `src/pflow/core/cache_analysis/CLAUDE.md` | Cross-boundary rec phrasing updated. |

### Agent-facing content

| File | Change |
|---|---|
| `src/pflow/guide/features/prompt-caching.md` | Two new sections: "Order chunks stable-to-volatile" and "Anthropic via proxies: per-chunk caching only fires with the `anthropic/` prefix". Rewritten twice to drop pflow-internal vocabulary. |
| `examples/core/prompt-caching-multi-chunk.pflow.md` | New fixture: 3 cache chunks declared stable-to-volatile, consumed by 2 LLM nodes. Validates clean. |

### Tests

| File | Change |
|---|---|
| `tests/test_core/test_cache_render.py` | New `TestComputeMarkerChunkIndices` class (15 tests). |
| `tests/test_core/test_llm_capabilities.py` | New `TestGetBreakpointBudget` class (5 tests). |
| `tests/test_nodes/test_llm/test_prompt_cache_rendering.py` | New `TestMultiBreakpointPlacement` class (10 tests). Rewrote existing `test_multi_chunk_declaration_order_preserved`. Updated header docstring. |
| `tests/test_nodes/test_llm/test_prompt_cache_below_min_runtime.py` | New `TestStripBelowMinSuppression` class (4 tests). |
| `tests/test_core/test_trace_report.py` | New `test_cached_system_renders_all_multi_breakpoint_markers` smoke test. |
| `tests/test_core/test_cache_analysis_per_id_emission.py` | Docstring updated (test logic unchanged). |

## Critical architectural insights (verified during implementation)

### 1. Hash side is marker-agnostic by design

The single most load-bearing fact about this change:

`runtime/engine/plan_node.py::_render_cache_for_hash` (lines 152-192) produces
`[{"name", "prose", "value"}, ...]` dicts — **no marker info, no provider
detection, no TTL, no `cache_control` field**. Markers are a pure
wire-format concern that affects what gets sent to the provider but NOT
what pflow hashes for memo cache identity.

This means **`compute_marker_chunk_indices` is called from the prep side
only**, never from the hash side. As a direct consequence:

- `_render_cache_for_hash` required ZERO changes.
- `compute_node_config` signature unchanged.
- `tests/test_runtime/fixtures/golden_config_hashes.json` did not drift.
- Existing memo cache entries remain valid across the upgrade.
- DD#19 byte-identity invariant preserved automatically.

The byte-equivalence tests (`test_hash_render_and_prep_render_byte_equivalent_*`)
compare `hash["prose"] + hash["value"]` vs `prep_block["text"]` —
multi-marker adds `cache_control` keys but does not touch `text` bytes,
so these tests pass unchanged.

### 2. `cache_ctx.prewarm` is the post-pre-flight truth

`_build_system_blocks` reads `cache_ctx.prewarm` (the engine pre-strips
this to `False` when `_should_disable_below_min_prewarm` fires on a
small batch prefix). This is the right input — using `config.prewarm`
or the IR-time prewarm would leak pre-strip state and over-reserve the
budget.

There's a documented sub-optimal corner: when prewarm passes pre-flight
but bails out at runtime (e.g., images present, alignment fail —
7 gate conditions in `llm.py:504-525`), `cache_ctx.prewarm` stays `True`
and we conservatively reserve a slot we won't use. Cost: one declared
marker we could have placed but didn't — strictly less aggressive
caching, never an API error. The 4-marker cap is never exceeded.

This is captured in `test_prewarm_declared_but_bailed_out_still_reserves_slot`.

### 3. The strip walker uses cross-channel cumulative tokens

`_strip_below_min_cache_markers` accumulates tokens across both channels
in order (`system_blocks` → `user_message_blocks`). A marker in the
prewarm channel sees cumulative tokens from EVERY block before it,
including all of `system_blocks`.

Consequence for the new suppression logic: "declared survives + prewarm
stripped" is structurally rare. Once `system_blocks` accumulates past
threshold, no later `user_message_blocks` marker can ever strip. This
is why the planned test under that name was reframed to
`test_prewarm_channel_suppresses_when_any_marker_survives` covering the
same suppression rule on a different, achievable channel config.

### 4. Single-chunk behavior is byte-identical

For 1 rendered chunk on any provider, `compute_marker_chunk_indices`
returns `(0,)`, producing exactly the same `cache_control` placement
as the v1 single-marker code. So:
- 1-chunk workflows are completely unchanged.
- All existing single-chunk tests in the suite pass unchanged.
- All single-chunk assertions like `sent[-1]["cache_control"] == {...}`
  remain correct.

The only test that needed an assertion change was the one explicitly
testing 3-chunk Anthropic placement.

## Deviations from the plan

### Intentional improvements

#### `assert` → `raise ValueError`

The plan specified using `assert n_rendered_chunks >= 1` to enforce the
caller contract. Ruff's `S101` rule flagged this as a security/style
violation (asserts disappear under `-O` flag in production Python).
Switched to:

```python
if n_rendered_chunks < 1:
    raise ValueError(
        "compute_marker_chunk_indices requires n_rendered_chunks >= 1 — "
        "the caller (_build_system_blocks) must guard the empty list."
    )
```

Same contract enforcement, lint-clean. The test
`test_raises_on_empty_rendered` was updated to expect `ValueError`
(with `match="n_rendered_chunks >= 1"`) instead of `AssertionError`.

#### Dropped two redundant tests

The plan listed `test_empty_returns_empty` and `test_empty_with_prewarm`
asserting an empty-tuple return for `n_rendered_chunks=0`. These are
inconsistent with the contract enforcement the SAME plan mandates (a
defensive empty return is exactly what the contract was meant to
prevent). Replaced with one positive test confirming the raise.
Coverage is equivalent.

#### Reframed `test_prewarm_channel_independent_of_declared_channel`

The plan called for a test where "declared survives + prewarm
stripped". The strip walker's cross-channel cumulative measurement
makes this structurally unreachable. Renamed to
`test_prewarm_channel_suppresses_when_any_marker_survives` and tested
the same suppression rule on a prewarm-only multi-marker configuration
(early below threshold, terminal above). Comments in the test
document the structural-rarity reason for the renaming.

### Quality fixes after first pass

#### Agent-facing prose: rewritten to drop internal vocabulary

First-pass guide additions used phrases like "multi-breakpoint
placement", "terminal marker", and "single-terminal-marker behavior" —
these are pflow-internal jargon. A fresh CLI agent author reading them
would have no anchor.

Rewritten to address agents who use the CLI to build workflows (not
contributors reading source):

- "Multi-breakpoint placement" → "per-chunk caching" / "each chunk's
  cache covers everything declared before it".
- "Terminal marker" → "the end of the full cached prefix remains
  cacheable as a single block".
- "Single-terminal-marker behavior" → "as a single cached prefix, not
  per-chunk".
- "Set the relevant API base via your provider configuration instead"
  (vague) → "address the model directly with `model: anthropic/claude-...`
  and provide `ANTHROPIC_API_KEY`" (concrete).

The example fixture's H1 description was similarly cleaned ("Demonstrates
multi-breakpoint prompt caching" → "Declares three `## Cache` chunks
ordered stable-to-volatile").

#### Fixture cache-chunk mis-parse

Initial example workflow had `Session context for ${session_id}:` inside
the `## Cache` block prose. The parser treated `${session_id}` as its
own chunk, yielding 4 declared chunks with `session_id` unused. Fixed
by removing the inline reference; `session_id` flows only through the
shell node, not the cache.

#### Monkeypatch tests refactored

Two suppression tests originally used `pytest.MonkeyPatch()` directly
with `try/finally monkey.undo()` boilerplate because they needed
per-text-value token counts (not constants). After review, switched to
the standard `monkeypatch: pytest.MonkeyPatch` fixture parameter and
called `monkeypatch.setattr` / `_stub_token_counter` directly. Matches
the style of every other test in the file.

## Test counts at each gate

| Gate | Tests | Result |
|---|---|---|
| `TestComputeMarkerChunkIndices` | 15 | All pass |
| `TestGetBreakpointBudget` | 5 | All pass |
| `TestMultiBreakpointPlacement` | 10 | All pass |
| `TestStripBelowMinSuppression` | 4 | All pass |
| `test_cached_system_renders_all_multi_breakpoint_markers` | 1 | Pass |
| `test_prompt_cache_rendering.py` (full) | 62 | All pass |
| `test_prompt_cache_below_min_runtime.py` (full) | 16 | All pass |
| `test_prompt_cache_hash.py` (golden baseline) | 16 | All pass unchanged |
| `test_cache_analysis_renderers.py` + `test_cache_analysis_analyze.py` | 377 | All pass |
| `make check` | — | Clean |
| `make test` (full suite) | 6998 passed, 1 skipped | Clean |

## Algorithm cheat-sheet (recorded for future readers)

```
n = rendered chunk count (post-ABSENT-filter)
budget = 4 if provider == "anthropic" else 1
if prewarm_consumes_slot: budget -= 1

if budget <= 1:           return (n - 1,)                  # terminal only
if n <= budget:           return tuple(range(n))           # all individual
                          return (*range(budget - 1),      # first (budget-1)
                                  n - 1)                    # plus terminal
```

Worked examples:
- Anthropic, 3 chunks, no prewarm → `(0, 1, 2)` — every chunk.
- Anthropic, 4 chunks, no prewarm → `(0, 1, 2, 3)` — every chunk.
- Anthropic, 5 chunks, no prewarm → `(0, 1, 2, 4)` — first three + terminal (chunk 3 merged in).
- Anthropic, 7 chunks, no prewarm → `(0, 1, 2, 6)` — first three + terminal.
- Anthropic, 5 chunks, prewarm=True (budget=3) → `(0, 1, 4)`.
- Anthropic, 1 chunk, no prewarm → `(0,)`.
- OpenAI/Gemini, any N → `(N-1,)` — terminal only.
- OpenAI, 3 chunks, prewarm=True (budget collapses to 0, then `<= 1` branch) → `(2,)` — still terminal.

## Loose ends (intentionally skipped, documented for follow-up)

### Operational verification (requires live API)

These were called out in the plan as "recommended" sanity checks but
need credentials to run:

1. **Real Anthropic smoke**: `uv run pflow examples/core/prompt-caching-multi-chunk.pflow.md` against a live Anthropic API key, then `grep '"cache_control"' ~/.pflow/debug/workflow-trace-*.json | wc -l` expecting `>= 3`. The unit + integration tests cover the placement algorithm thoroughly, but no automated test exercises the full live wire format.

2. **`pflow analyze-cache` before/after diff**: compare analyzer JSON for a known baseline workflow before vs. after the change. The 377 cache_analysis regression tests provide programmatic coverage; the explicit JSON diff was not run.

### Out-of-scope items for separate GH issues (per user instruction, deferred)

The plan's "Out of Scope / Deferred to GH Issues" section listed 7
follow-up items. User explicitly said to handle GH issue creation
separately:

1. Per-marker telemetry in `pflow report` (derive per-marker
   hit/miss from `cache_read_input_tokens`).
2. Cross-node analyzer recommendations (recommend chunk reorderings
   based on shared-prefix patterns).
3. Routed-Anthropic diagnostic (`cache.routed-provider-degraded`
   when an Anthropic-looking model name routes through a proxy
   prefix). Currently covered by a guide doc note only.
4. Convention enforcement (advisory based on observed chunk-change
   frequencies from trace history).
5. Budget exhaustion advisory (`cache.too-many-chunks-for-provider`
   INFO when `n_rendered_chunks > budget`).
6. Per-chunk TTL override (mix 1h stable prefix + 5m volatile tail).
7. Per-marker analyzer threshold modeling (predict per-EARLIER-marker
   activation rates).

## Pre-implementation audit (recorded for reproducibility)

Per the plan, ran the stale-vocabulary grep BEFORE writing code:

```
$ grep -rn 'single-breakpoint\|LAST chunk only\|LAST block only\|single cache_control\|single provider cache breakpoint\|DD#11' src/ tests/ architecture/ docs/
src/pflow/core/llm_client.py:236-237        single-marker contract comment
src/pflow/core/cache_render.py:183          _build_cache_control_marker docstring
src/pflow/core/cache_analysis/CLAUDE.md:184 cross-boundary recs phrasing
src/pflow/nodes/llm/llm.py:790-791          _build_system_blocks docstring
tests/test_nodes/test_llm/test_prompt_cache_rendering.py:1   header docstring
tests/test_nodes/test_llm/test_prompt_cache_rendering.py:208 inline comment
```

Plus `cache_analysis/analyze.py:6171` and
`tests/test_core/test_cache_analysis_per_id_emission.py:2868` flagged
by the plan directly. All 8 sites updated.

Post-implementation re-grep returns zero hits for any stale phrase.

JSON fixture audit (`grep -rn '"cache_control"' tests/fixtures/ examples/`)
returned zero hits — no committed JSON files embed wire-format markers
that would drift under multi-marker placement.

## Risk register (final state)

| Risk | Status | Evidence |
|---|---|---|
| Hash drift breaking memo cache | Prevented | Hash side untouched. `test_golden_baseline_hashes_match` passes unchanged. |
| 4-marker API limit exceeded | Prevented | `compute_marker_chunk_indices` enforces budget. Prewarm conservatively reserves slot. Both tested. |
| Non-Anthropic providers break | Prevented | Algorithm collapses to single terminal marker for budget ≤ 1. Tested for openai/gemini/None/unknown. |
| Cross-workflow SUM in analyzer breaks | Prevented | Markers cover contiguous prefixes in declaration order; per-call denominator unchanged. 377/377 analyzer tests pass. |
| `_strip_below_min_cache_markers` fails | Prevented | Function was already plural-aware; suppression added without changing strip behavior. 16/16 strip tests pass. |
| Warning noise on multi-chunk workflows | Mitigated | Suppression-when-any-marker-survives. Locked by `TestStripBelowMinSuppression`. |
| OpenRouter/Bedrock/Vertex silent degradation | Documented | Guide section with concrete fix. Runtime diagnostic deferred to GH follow-up. |
| Stable-to-volatile convention not enforced | Documented | Guide explains; multi-marker is never net-negative vs single-marker. |
| Heterogeneous batch with prewarm — sub-optimal per-item placement | Documented | Conservative slot reservation never exceeds API cap. `_build_system_blocks` comment + dedicated test. |
| Stale docstrings (8 sites) | Resolved | All updated in lockstep with code change. |
| Trace fixture drift | Prevented | Audit returned zero hits. |

## Reference for future readers

- Algorithm: `src/pflow/core/cache_render.py::compute_marker_chunk_indices`
- Budget table: `src/pflow/core/llm_capabilities.py::get_breakpoint_budget`
- Placement loop: `src/pflow/nodes/llm/llm.py::_build_system_blocks` (chunk_block_offset + marker_indices loop)
- Strip suppression: `src/pflow/nodes/llm/llm.py::_strip_below_min_cache_markers` (per-channel "any surviving marker" check)
- Test matrix: `tests/test_core/test_cache_render.py::TestComputeMarkerChunkIndices`
- End-to-end placement tests: `tests/test_nodes/test_llm/test_prompt_cache_rendering.py::TestMultiBreakpointPlacement`
- Strip-suppression unit tests: `tests/test_nodes/test_llm/test_prompt_cache_below_min_runtime.py::TestStripBelowMinSuppression`
- Agent-facing guide: `src/pflow/guide/features/prompt-caching.md` — sections "Order chunks stable-to-volatile" and "Anthropic via proxies"
- Example fixture: `examples/core/prompt-caching-multi-chunk.pflow.md`

## Anthropic documentation reference (verified at implementation time)

Source: https://platform.claude.com/docs/en/docs/build-with-claude/prompt-caching

Key facts:
- Max 4 `cache_control` breakpoints per request (API returns 400 if exceeded).
- Pricing is delta-based on writes — adding more breakpoints does NOT cost more.
- Reads: longest matching prefix wins via a 20-block lookback window.
- Min tokens applies per-breakpoint cumulatively from prompt start (matches existing `_strip_below_min_cache_markers` semantics).

So multiple breakpoints are strictly an improvement: the write side is
free, and the read side gains independence (a change in chunk[N] no
longer invalidates the cached portion through chunk[N-1]).

---

# Follow-up: code review pass (3 specialist agents + 1 verifier)

After the initial commit landed, ran `/code-review` with the 3 most relevant
agents in parallel: **review-impact-completeness**, **review-silent-failures**,
**review-feature-interactions**. A targeted **pflow-codebase-searcher** agent
verified two analyzer-side claims before classification.

## Verdicts

| Class | Count | Notes |
|---|---|---|
| Critical | 0 | No correctness bugs found. |
| Confirmed (action taken) | 6 | All addressed in the follow-up commit. |
| Confirmed but deferred | 3 | GH issue work, per user instruction to handle separately. |
| Disputed | 4 | See below for evidence. |

## Disputed findings (with evidence)

**Cost projection under-estimated 4× under multi-marker** — feature-interactions
claimed Anthropic charges per-write per scope so 4 markers = 4 × cache_creation
cost. Verifier confirmed `core/cache_analysis/cost_estimation.py:281-318, 494-593`
uses `row.cache_active.tokens_estimated` (cumulative-through-terminal) once
per call with a single `write_rate × tokens` math. This matches Anthropic's
documented delta-pricing — writing markers at [0..K] and [0..N] in the same
request charges deltas, not duplicates. The reviewer mis-modeled the pricing.

**Parallel batch heterogeneous ABSENT × OpenAI sticky routing** — flagged
as a multi-marker regression. Pre-existing behavior (ABSENT-filter and
per-item rendering existed before this commit). Not introduced.

**`cache_skipped_reason` partial-survival** — reviewer noted "This is actually
correct behavior" inline; the observability gap is real but is covered by
deferred follow-up #410 (per-marker telemetry).

**Trace report `default=str` for hypothetical nested markers** — markers
today are flat. The implementation comment already documents the migration
trigger to `copy.deepcopy`. Adding a test for a hypothetical state isn't
useful; the new deepcopy-aliasing regression test (item #1 below) covers
the migration trigger.

## Follow-up commit content

Six in-commit fixes (3 tests, 3 docstring/comment touch-ups):

1. **`test_mutating_one_markers_top_level_does_not_leak_to_others`** — behavioral
   guard for the shallow-copy footgun documented at `_build_system_blocks:872-875`.
   Today markers are flat, so this tests the simpler top-level case; the
   inline comment documents the deepcopy migration trigger for nested markers.

2. **Centralization docstrings tightened across 3 sites**:
   - `_compute_marker_chunk_indices`: added "Scope of centralized placement"
     block explaining `_build_user_message_blocks` is a parallel emitter by
     construction; the 4-marker cap is shared via `prewarm_consumes_slot`.
   - `_build_user_message_blocks`: inline cross-reference at the
     `cache_control` emit site.
   - `_build_openai_cache_kwargs`: replaced "last block" framing with the
     explicit budget-1 + `_compute_marker_chunk_indices` derivation.

3. **`_strip_below_min_cache_markers` docstring**: added "Asymmetry to be
   aware of" + "`*_measured_tokens` semantics" paragraphs documenting the
   cross-channel cumulative vs per-channel suppression interaction, and the
   smallest-stripped-scope semantics of the measured tokens field.

4. **`_compute_marker_chunk_indices` body**: inline comment on the `budget <= 1`
   guard explaining the budget=0 hypothetical (when a future budget=1 provider
   also debits a prewarm slot — still emits terminal marker; revisit if a
   future provider has a strict shared cap).

5. **`test_cross_channel_cumulative_prewarm_marker_survives_via_system_accumulation`**
   — locks the cross-channel cumulative + per-channel suppression asymmetry
   newly documented in #3 above. 3 declared (cum 400/800/1200, threshold 1024,
   so 0,1 strip and 2 survives) + prewarm marker (cum 1250, survives because
   the system-channel already crossed threshold).

6. **`test_openai_prompt_cache_key_pinned_hash_for_known_content`** — pins
   the OpenAI MD5 hex (`180791359f4acd84630ff34114e4bf39`) for known cache
   content. Trips loudly if `get_breakpoint_budget("openai")`,
   `_build_cache_control_marker`, `_compute_marker_chunk_indices` for
   budget=1, or the deterministic-JSON serialization changes in a way that
   would silently invalidate sticky routing across deploys.

## Additional cleanup (after follow-up commit)

Three more touch-ups landed after the review pass:

**A. Function renamed: `compute_marker_chunk_indices` → `_compute_marker_chunk_indices`.**
Aligns with the project convention for one-call-site private helpers in
`cache_render.py` (`_resolve_chunk_value`, `_build_cache_control_marker`,
`_resolve_static_prefix_for_cache`). Signals the private-contract intent —
future callers can't sneak in without acknowledging the empty-input guard.

**B. End-to-end `WorkflowRunner` test** at
`tests/test_integration/test_prompt_cache_multi_chunk.py`. Drives the
committed `examples/core/prompt-caching-multi-chunk.pflow.md` through
`WorkflowRunner.run()` with the mock LLM and asserts that both consumer
LLM nodes (`summarize`, `translate`) reach the adapter with 3 system blocks,
each carrying a `cache_control` marker. Closes the cross-layer test gap
that unit-level placement tests don't cover.

**C. This addendum.**

## Confirmed but deferred (filed for separate GH follow-up)

1. **Routed-Anthropic silent degradation** — both reviewers flagged. Currently
   covered by the guide doc; deferred to GH #409 per the commit message.
   Real user-visible silent failure: a workflow with `model:
   openrouter/anthropic/claude-sonnet-4-5` and 4 declared chunks silently
   gets 1 marker. Fix would emit an INFO diagnostic
   `cache.routed-provider-degraded` from `_compute_marker_chunk_indices` when
   `provider_name is None AND n_rendered_chunks > 1 AND bare_model matches
   /anthropic|claude-/`. ~20 LOC + 3-4 tests + warning-catalog entry.

2. **`_attribute_root_cause` priority under multi-marker.** Verifier confirmed
   the function reads ONLY `chunks_skipped` (`analyze.py:7983-8008`); zero-cache
   telemetry and below_min are not consulted. When suppressed-below-min +
   ABSENT chunks coexist, the analyzer attributes to ABSENT branches —
   pointing the user at the wrong root cause. Pre-existing design choice;
   multi-marker just makes "suppressed below_min + ABSENT" the more common
   case. Worth filing as analyzer follow-up.

3. **Storage-mode-shared × parallel batch × `## Cache` UNSUPPORTED COMBO.**
   Pre-existing silent-but-benign behavior documented in `runtime/CLAUDE.md`.
   Multi-marker widens the corrupted-state surface (N possible marker
   topologies vs. just N positions). No current consumer reads parent's
   `__pflow_cache_render__` after a parallel batch, so the corruption is
   unobservable today. File the runtime guard before Task 160's analyzer
   refactor lands and might add such a consumer.

## Final state — verification

Both commits combined:

| Gate | Result |
|---|---|
| `make check` | clean (ruff/ruff-format/mypy/deptry) |
| Targeted tests (cache_render, capabilities, hash, rendering, below-min, batch, multi-chunk integration) | 198/198 pass |
| `make test` (full suite) | 7002 passed, 1 skipped |
| Pre-implementation grep audit (stale single-marker phrases) | zero hits |
| JSON fixture audit (`'"cache_control"'` in committed fixtures) | zero hits |
| Hash byte-identity (`test_golden_baseline_hashes_match`) | pass unchanged |

## Test counts — running total

Initial commit (76334d0): +39 new tests across 5 files.
Follow-up commit (review pass fixes): +3 tests (deepcopy-aliasing,
cross-channel cumulative, OpenAI MD5 pin).
This cleanup pass: +1 end-to-end test through `WorkflowRunner`.

Total new test coverage attributable to multi-breakpoint work: **43 tests**.

## Lessons recorded for future readers

1. **The "fully centralized" framing was technically false.** `_build_user_message_blocks`
   is a parallel `cache_control` emission site (single-marker by construction).
   Both reviewer + this addendum now document the constraint explicitly.

2. **One reviewer was confidently wrong about Anthropic's pricing semantics.**
   The cost-projection 4×-under-estimate claim was a categorical error
   contradicting both the implementation plan's pricing citation and the
   actual Anthropic docs. Verification before classification caught this —
   trusting the reviewer's verdict on the surface would have triggered a
   wrong analyzer-refactor.

3. **The deepcopy comment was paper without a test.** The reviewer was
   right: a single behavioral test (mutate one marker, assert neighbors
   unchanged) makes the migration trigger enforceable.

4. **The strip walker's cross-channel cumulative vs per-channel suppression
   asymmetry is subtle.** Documenting it explicitly in the function docstring
   prevents future contributors from "fixing" the cumulative to reset
   per-channel — which would silently break Anthropic's documented
   "cumulative through preceding content" semantics.

---

# Follow-up 2: severity-aware status + routed-Anthropic advisory (D)

Commit: `89752332 feat: severity-aware status + routed-Anthropic cache advisory [skip review]`

After the review pass and the cleanup commit, one deferred item from
the review remained substantive enough to investigate: **D — emit a
runtime diagnostic when a workflow uses a routed-Anthropic model
identifier with multi-chunk caching, so the silent degradation
(per-chunk reuse lost, only terminal cache fires) stops being
asymptomatic.**

## The trap I almost walked into

My first instinct was to ship D as a `Severity.WARNING` catalog entry
emitted from `_build_system_blocks`. The investigation surfaced a
specific concern: `_determine_status` (`runner.py:557-560`) was
**severity-blind** — any entry in `__warnings__` flipped DEGRADED
regardless of catalog severity. Routed-Anthropic users (especially
compliance-routed callers stuck on Bedrock/Vertex for AWS-only /
GCP-only deployments) would suddenly start seeing DEGRADED every run,
with no silence mechanism. The "fix" was telling them to either
abandon caching or live with permanent CI-gate failures — neither is
acceptable.

I considered three workarounds:

- **C1 — analyzer-only emission**: split the same logical concern
  across two emission paths. Adds a "advisories live in analyzer OR
  runtime depending on tier" surface that ages badly.
- **C2 — per-node IR silence flag** (e.g.,
  `cache_silence: ["cache.routed-provider-degraded"]`): new IR field
  just to manage one warning. Top-10% codebases don't add silence
  flags; they make the signal itself correct.
- **C3 — INFO + accept DEGRADED**: harshest on compliance-routed
  users; they get no escape.

None of these are the right answer. The actual problem was upstream:
**the project today has no way to express "this is informational"**
because `_determine_status` treats every `__warnings__` entry as a
regression signal. The right fix is to make status determination
read the severity the catalog already encodes.

The user's reframing nailed it: *"prioritize simplicity of the FINAL
code, not how easy it is to get there. When in doubt we should ask
ourselves what's the right solution that the top 10% of codebases
similar to this one would implement."* A top-10% codebase has a
severity-aware status determination, not a workaround layer for
specific catalog IDs.

## Scout before commit

Before adopting the severity-aware status fix I scouted the breakage
surface to verify the change wasn't a behavior regression for
shipped workflows:

- **12 INFO catalog entries exist** (more than my initial guess of
  3-4). All are analyzer-only emissions today.
- **22 test files touch INFO catalog IDs** in the catalog string
  sense; none assert DEGRADED for INFO emissions.
- **Tests that assert `WorkflowStatus.DEGRADED` AND mention an INFO
  catalog ID**: grep intersection returned zero matches.
- Verified: every shipped runtime emitter that writes to
  `__warnings__` writes `Severity.WARNING` (or a legacy string/dict).
  Zero existing test breakage.

The conclusion: the severity-aware fix is forward-looking with no
behavior shift for shipped workflows. It makes the abstraction
correct so every future INFO entry gets the right semantics for free.

## The shape that landed

**Two changes shipped together as one coupled commit:**

### Change A — severity-aware `_determine_status`

```python
def _is_degrading_warning(value: Any) -> bool:
    """Severity.INFO Diagnostics are advisories; everything else degrades."""
    if isinstance(value, Diagnostic):
        return value.severity is not Severity.INFO
    return True  # legacy string/dict shapes preserve pre-existing behavior
```

`_determine_status` now filters `__warnings__` through this helper.
INFO Diagnostics surface in reports and `result.warnings` but don't
flip status. WARNING/ERROR diagnostics still degrade. Legacy
string/dict warning shapes default to degrading — preserves
pre-catalog behavior for the LLM adapter, on-error recovery, and
every other producer that hasn't migrated yet.

10 LOC core change. 9 dedicated unit tests covering the matrix
(success/info/warning/error/mixed/legacy-string/legacy-dict/
template-errors/error-action override).

### Change B — `cache.routed-provider-degraded` INFO advisory

Built on Change A's foundation:

- **Detection helper** `_looks_like_routed_anthropic(model)` in
  `cache_render.py`: `detect_provider(model) is None` AND lowercased
  model contains `claude` or `anthropic`. One substring check after
  the provider consult — no router enumeration. Catches OpenRouter,
  Bedrock, Vertex, and any future routing prefix containing those
  substrings.
- **Catalog entry** in `warning_catalog.py`: `Severity.INFO`,
  `CACHE_ADVISORY_CATEGORY`, priority 30 (alongside
  `cache.below-min-observed`). Required context:
  `(node_id, model, n_rendered_chunks)`. Concrete remediation
  suggestion: switch to `anthropic/` + `ANTHROPIC_API_KEY` for
  per-chunk caching, or knowingly accept the limitation for
  compliance routing.
- **Runtime emitter** `_emit_routed_provider_degraded_advisory` in
  `llm.py`: called from `_build_system_blocks` after marker placement
  when `len(rendered) > 1 AND _looks_like_routed_anthropic(model)`.
  `setdefault` semantics so authoritative warnings
  (`cache.below-min-rendered`, `cache.prewarm-disabled-below-min`)
  take precedence when both would fire.

## Why this is the simplest final shape

| Option | New surfaces | Reasoning load on future contributors |
|---|---|---|
| Analyzer-only | 0 | "Some advisories live in analyzer, others in runtime" |
| Per-node silence flag | New IR field, lookup, docs | "Catalog IDs can be silenced per-node" |
| Runtime WARNING + DEGRADED | 0 | "INFO entries flip status" (relitigate per entry) |
| **Fix `_determine_status` + ship D as INFO** | 0 | None — `Severity.INFO` already means what its name says |

A new contributor adding the 13th INFO catalog entry tomorrow gets
the right semantics without reading this conversation. The catalog
mechanism stays uniform; status determination stops throwing away
information it had.

## Trigger conditions (locked by tests)

The advisory fires ONLY when ALL these hold simultaneously:

1. Workflow declares `## Cache` with 2+ chunks.
2. A node opts in via `prompt_cache: [a, b, ...]` with 2+ entries
   that resolve (not all ABSENT-filtered).
3. `detect_provider(model)` returns None (model identifier doesn't
   start with `anthropic/`, `openai/`, `gemini/`, or bare
   `claude-*`/`gpt-*`/`o1`/`o3`/`o4`/`gemini-*`).
4. Lowercased model identifier contains `claude` or `anthropic` as a
   substring.

Concrete fires:
- `model: openrouter/anthropic/claude-sonnet-4-5` + 3 chunks
- `model: bedrock/anthropic.claude-sonnet-4-5-v1:0` + 2 chunks
- `model: vertex_ai/claude-sonnet-4-5@20250514` + 2 chunks

Concrete doesn't-fire:
- `model: anthropic/claude-...` (native — already gets multi-marker)
- `model: claude-...` bare (native — already gets multi-marker)
- `model: openrouter/openai/gpt-4o` (no `claude`/`anthropic`)
- `model: ollama/llama-3` (unknown but not Anthropic-flavored)
- Routed-Anthropic + single-chunk (no degradation to surface)

## Catalog lock-test surface

Adding a new catalog ID requires touching 4 lock-tests in
lockstep — this is the catalog's intentional design pressure so
nothing ships without docstring/coverage discipline:

1. `test_cache_analysis_warnings.py::test_catalog_size_matches_v1_inventory`
   — count assertion 30 → 31 + entry in the inventory enumeration.
2. `test_cache_analysis_warnings.py::test_every_id_round_trips_through_make_diagnostic`
   — kwargs sample for `make_diagnostic(...)` round-trip.
3. `test_cache_analysis_per_id_coverage.py::test_every_catalog_id_has_a_kwargs_sample`
   — kwargs sample for the per-id coverage.
4. `test_cache_analysis_per_id_coverage.py::test_emitted_diagnostics_round_trip_for_real_producer_paths`
   — producer-driven round-trip (drives the actual emitter).
5. `test_mcp_server/test_analyze_cache_tool.py::test_docstring_lists_every_catalog_id`
   — docstring of the MCP `analyze_cache` tool lists every catalog ID.

Also updated in lockstep (not lock-tested but conventionally synced):
- `cache_analysis/CLAUDE.md` count reference (30 → 31).
- `cache_analysis/CLAUDE.md` paragraph documenting the new
  ID's runtime-emission + INFO-severity choice.
- `guide/features/prompt-caching.md`: one-line note telling agents
  the advisory exists so they don't have to dig.
- MCP `analyze_cache` tool docstring: new ID added to the catalog
  list.

## Verification

| Gate | Result |
|---|---|
| `make check` | clean (ruff, ruff-format, mypy, deptry) |
| Targeted (severity + detection + emission + lock-tests) | 30/30 new tests pass |
| `make test` (full suite) | 7032 passed (+30), 1 skipped |
| Pre-commit hooks | clean (commit hook ran ruff + ruff-format, both passed) |

## Total test coverage attributable to multi-breakpoint work

| Commit | New tests |
|---|---|
| Initial (76334d0e) | +39 |
| Review fixes (d53ef0e3) | +4 |
| Severity + routed-Anthropic (89752332) | +30 |
| **Total** | **+73** |

## Lessons recorded

5. **A reasoning trap was lurking in `_determine_status`.** The
   abstraction was "warnings exist → degraded," which conflated
   advisory signals with regression signals. The catalog had typed
   severities all along; the status code never read them. Easy to
   miss because every existing producer happened to align with the
   conflated semantics by accident.

6. **Top-10% framing is a useful frame, but it must point at
   "simpler final code" not "more sophisticated final code."** My
   first three options (analyzer-only, silence flag, INFO+DEGRADED)
   were all more sophisticated than the upstream fix. The user's
   reframing cut through it cleanly: ask what the catalog's names
   already mean, then make the code match. No new surface needed.

7. **Scout before adopting a behavior-change refactor.** I assumed
   the severity-aware change would break tests. The grep intersection
   returned zero matches because every shipped INFO emitter is
   analyzer-only, not runtime. 5 minutes of scouting saved the
   "should I really change this?" loop.

8. **The catalog lock-test surface is intentional friction.** 5 files
   needed updates for one new entry. That's the design — every new
   ID has to prove its docstring, kwargs sample, producer round-trip,
   and MCP exposure in lockstep, or one of the lock-tests fires.
   Adding catalog entries is friction-worth-paying.

Issue #458, opened 2026-06-01
`--only`: warn when a changed input feeds a frozen upstream node (silent stale-input under snapshot)


spinje:
## Summary

After the `--only` snapshot-semantics change (#443), `--only <node>` runs the target against a **frozen snapshot** of the most recent full run — upstream nodes are restored from the trace, never re-executed. This is correct and intended.

The gap: when a user/agent **passes a changed input** on an `--only` run, the new value reaches only the **target** (and any node that actually executes). Frozen upstream nodes are *not* re-derived from it, so any upstream output the target consumes still reflects the **old** input — **silently**. There is no warning, unlike the sibling degraded-snapshot case which fires a loud `⚠`.

This is a documented limitation (see `context/adr/0002-443-only-snapshot-source.md` → Limitations, "Frozen upstream + live inputs are an incoherent mix"), but it's a real foot-gun for AI agents, who may pass an input out of habit and trust that it propagated.

**This issue is a follow-up to #443, not a re-open.** #443 was about side-effect *re-firing*; this is about *input-coherence detection*. It is purely additive — a safety advisory — and must **not** change `--only`'s execution semantics or weaken the no-re-fire guarantee.

## Reproduction (verified against current `main` post-#443)

Workflow `inp.pflow.md`:

```markdown
# Input + --only

## Inputs

### tag

A tag used by BOTH the upstream node and the target.

- type: string
- required: false
- default: DFLT

## Steps

### up

Upstream: bakes the input into its output.

- type: shell

\`\`\`shell command
echo "up-${tag}"
\`\`\`

### target

The --only target: reads the input DIRECTLY and the upstream's output.

- type: shell

\`\`\`shell command
echo "target tag=${tag} up=${up.stdout}"
\`\`\`
```

```console
$ pflow inp.pflow.md tag=A                 # full run
target tag=A up=up-A

$ pflow inp.pflow.md --only target tag=B   # exit 0, no warning
target tag=B up=up-A
#              ^^^^ STALE: `up` was NOT re-run with tag=B; its output is frozen at the tag=A value
```

The target's *direct* `${tag}` correctly becomes `B`; but `${up.stdout}` is the frozen `up-A`. The result is silently incoherent (`tag=B` + `up-A`).

## Why this happens (mechanism — for the implementer)

- The CLI param is seeded into the shared store and **is** used (`shared["tag"] = "B"`). It is not ignored.
[...]

Issue #450, opened 2026-05-29
Empty-input batch reports DEGRADED with a misleading warning on legitimate empty iterations


spinje:
## Problem

When a batch node's input list resolves to empty, the workflow is reported as **DEGRADED** and the batch warning tells the user to "fix the upstream data." But an empty input list is the *normal terminal state* of the iteration patterns the guide now teaches:

- a backward-edge loop draining a queue — the final iterations are legitimately empty; and
- a batch over a filtered list that matched nothing.

So a fully-correct run reports DEGRADED and points the agent at data that isn't broken. For an agent-first tool this mis-signals on exactly the patterns we recommend — and the runtime can't distinguish "empty because the work is done" from "empty because upstream broke" (there is no loop construct to consult).

## Also in scope

- The distinction between a *legitimate* empty batch and a *genuine* anomaly is lost today. Items that ran but produced empty output, and per-item errors, are real signals that must keep degrading.
- Any fix must keep the **text and JSON/MCP surfaces consistent**. An advisory that's demoted in the text summary but still counted under `warnings` in JSON would just relocate the problem — an agent parsing `len(warnings)` would still see a clean run as warned.

## Docs gap (same area)

The branching guide's **Loops** section illustrates a `workflow`-node loop body by referencing `./process-chunk.pflow.md` — a child workflow that does not exist in the repo, so the example cannot be reproduced. Separately, the runnable Loops example has no regression guard, so it can silently rot to something broken that agents would copy.

## Acceptance criteria

- [ ] An empty-input batch completes as **SUCCESS** (exit 0), not DEGRADED.
- [ ] It surfaces a **non-degrading advisory** with accurate wording — names the legitimate case, no "fix the upstream data" misdirection.
- [ ] Genuine batch anomalies (items produced empty output; per-item errors) **still degrade**.
- [ ] **Text and JSON** output treat the advisory consistently — not counted as a warning on either surface.
- [ ] The branching-guide Loops section has **no dangling file reference**.
- [ ] The runnable Loops example is covered by a **regression test**.

Issue #338, opened 2026-04-21
`--only`: support dotted paths for targeting nodes inside sub-workflows


spinje:
## Problem

`--only` only resolves nodes at the top level of the orchestrator workflow. When the target node lives inside a sub-workflow (or deeper — a sub-sub-workflow), there's no way to reach it.

```bash
# This fails — write-lyrics is inside the create-songs sub-workflow
pflow orchestrator.pflow.md --only write-lyrics

# Error:
# Node 'write-lyrics' not found
# Available nodes: fetch-sources, analyze-sources, create-songs, evaluate-songs, ...
```

The only workaround is `--only create-songs`, which re-runs **every** node downstream of the cache boundary inside the sub-workflow — not just the target node. In our case, this means re-running 7 review sub-workflows and 2 rewrite steps we don't need, turning a ~$0.08 experiment into a ~$0.50+ one.

## Use case

We're A/B testing model quality on individual pipeline nodes (swapping Gemini Flash → Claude Sonnet on `write-lyrics`, then `creative-direction`, then `rewrite-emotional`, etc.). Each test changes the `- model:` param on one node, and we want to compare just that node's raw output against the baseline.

The pipeline structure is:

```
orchestrator.pflow.md
  → create-songs (batch sub-workflow × 4 items, parallel)
      → creative-direction (cached)
      → song-architecture (cached)
      → choose-chorus (sub-sub-workflow, cached)
      → write-lyrics          ← we want to target THIS
      → emotional-reviews     ← don't need these
      → rewrite-emotional     ← or these
      → craft-reviews
      → rewrite-craft
      → generate-suno-prompt
```

What we want:

```bash
# Run through write-lyrics inside create-songs, skip everything after it
pflow orchestrator.pflow.md --only create-songs.write-lyrics
```

This would: resolve upstream through the sub-workflow hierarchy, execute `create-songs` but stop after `write-lyrics` inside it, skip `emotional-reviews` through `generate-suno-prompt`.

## Proposed syntax

Dotted path: `--only <top-level-node>.<child-node>` (and deeper: `--only create-songs.choose-chorus.select-chorus`).

The path segments map to the sub-workflow nesting: each segment is a node ID at its level. The last segment is where execution stops.

### Setup

Any workflow with a sub-workflow containing multiple nodes. Minimal example:

**parent.pflow.md:**
```markdown
# Parent

## Steps

### step-a
- type: llm
- prompt: "Say hello"

### step-b
- type: workflow
[...]

PR #459, opened 2026-06-01
fix: `--only` re-runs against a frozen snapshot instead of re-firing upstream (#443)

### LINKED ISSUE #443: `--only` re-fires side-effecting upstream nodes after the cache-defaults flip
## Summary

`--only <node>` walks the graph from the start node and **executes every upstream node** until it reaches the target (then stops). For each upstream node it consults the memo cache: a cache hit returns instantly without running `exec()`; a cache miss runs the node for real.

After the cache-defaults flip (only `llm` caches by default; `shell`/`code`/`http`/`file`/`mcp`/`claude-code` default to `cache: false`), non-`llm` upstream nodes are **always a cache miss** under `--only`, so they **re-execute — and their side effects re-fire** on every `--only` invocation.

This changed `--only`'s behavior without anyone analyzing the interaction (it was a silent side effect of the cache flip, not a deliberate `--only` change).

## Before / after

Workflow: `fetch (shell: gh pr create) → summarize (llm)`. Agent iterates on the `summarize` prompt with `pflow wf.pflow.md --only summarize`.

| | Upstream `shell: gh pr create` under `--only summarize` |
|---|---|
| **Before the flip** | `cache_enabled=True` → memo hit → cached result returned, `exec()` not run → **PR not re-created** |
| **After the flip** | `cache_enabled=False` → memo skipped → `exec()` runs → **PR re-created every `--only` run** |

The functional result (target node gets fresh upstream inputs) is arguably *more* correct, but it defeats `--only`'s core purpose — cheaply iterating on one node without disturbing upstream — and silently re-triggers side effects.

## Why this matters

`--only` is the primary "tune one node without paying for the rest" tool. Its value depends on upstream being reused, not re-run. Post-flip that only holds when upstream is `llm` (or explicitly `cache: true`). For the common `shell`/`code`/`http` → `llm` shape, every `--only` run now re-executes the upstream side-effecting/external-state nodes. For side-effecting upstream (`gh pr create`, `git push`, an HTTP POST, an `mcp` write) this is a real foot-gun, not just a performance change.

This is the same class of problem the cache-defaults flip itself set out to fix (silent staleness vs. silent re-firing) — the flip closed the staleness hole in normal runs but opened a re-fire hole in `--only`.

## Current behavior (verified against code)

- `WorkflowEngine._run_inner` (`src/pflow/runtime/engine/engine.py`) sets `curr = workflow.start_node` and runs each node in a `while curr:` loop, breaking **after** executing the `--only` target. There is no "restore upstream from a prior trace" step — upstream is genuinely re-walked and re-executed.
- Per-node memo lookup is gated on `cache_enabled` (`runtime/engine/plan_node.py` read gate; `engine.py` write gate). With `cache_enabled=False`, the lookup is skipped and `exec()` runs.
- `cache_enabled` now defaults to `node_type == "llm"` via `_default_cache_for_node_type` (`src/pflow/runtime/compilation/compiler.py`), so all non-`llm` upstream nodes miss the cache under `--only`.

## Options to consider (not prescribing)

1. **Status quo + docs only** (done in the cache-defaults PR): the `--only` guidance across `pflow guide`, the CLI reference, and the MCP instructions now states that uncached upstream re-executes and warns about side-effecting upstream re-firing. Cheapest; the foot-gun persists.
2. **Restore uncached upstream from the last trace.** Under `--only`, instead of re-executing an uncached upstream node, hydrate its outputs from the most recent saved trace for this workflow (`~/.pflow/debug/`). Preserves `--only`'s cheap-iteration value regardless of node type and avoids re-firing side effects. Needs: trace lookup/selection, a staleness story (what if the file/inputs changed since the trace), and a fallback when no trace exists.
3. **`--only` reads the memo cache for *all* upstream nodes regardless of `cache_enabled`.** Treat `--only` as an explicit "I want cached upstream" mode. Simpler than trace restoration but reintroduces the staleness the flip removed (a cached `shell` result may not reflect current disk state) — only acceptable because the user explicitly opted into single-node mode.
4. **Require upstream to be cached and error otherwise.** If `--only`'s upstream contains an uncached node, fail with a clear message ("upstream node X doesn't cache; add `cache: true` or run the full workflow"). Loud instead of silently re-firing, but pushes work onto the author.

## Acceptance criteria (whichever option)

- Running `--only <node>` does not silently re-fire side effects of upstream nodes that the user expects to be skipped — or, if it must, it says so loudly before doing it.
- `--only`'s cheap-iteration purpose holds for the common `shell`/`code`/`http` → `llm` shape, OR the limitation is explicit and the alternative is named.
- The `--only` documentation (guide `core.md`, `nodes/llm.md`, CLI reference, MCP instructions) matches whatever lands.

## References

- `src/pflow/runtime/engine/engine.py` → `_run_inner` (the start-to-target walk; no trace restoration).
- `src/pflow/runtime/engine/plan_node.py` / `engine.py` → memo read/write gates on `cache_enabled`.
- `src/pflow/runtime/compilation/compiler.py` → `_default_cache_for_node_type` (the `node_type == "llm"` default).
- Docs already carrying the interim warning: `src/pflow/guide/core.md`, `src/pflow/guide/nodes/llm.md`, `docs/reference/cli/index.mdx`, `src/pflow/mcp_server/resources/instructions/mcp-agent-instructions.md`, `src/pflow/cli/CLAUDE.md`.
- Surfaced during the `feat/cache-defaults-iteration-docs` work (PR #442); related to the cache-defaults flip and #441.

spinje:
## Summary

`--only <node>` now runs the target against a **frozen snapshot** of the most recent full run instead of re-walking the graph. After the cache-defaults flip, `--only` re-executed every uncached upstream node, **silently re-firing their side effects** (e.g. `shell: gh pr create`, `git push`, an HTTP POST) on every iteration — defeating `--only`'s "tune one node cheaply" purpose. Now upstream is restored from the last full successful run's trace and **never re-run**; only the target executes.

Fixes #443

## Changes

- **Engine** (`runtime/engine/engine.py`): `--only` routes to `_run_only_snapshot` (seed snapshot → run only the target via `_execute_node`), and the graph walk's `--only` branches are deleted (the walk is now full-run-only). New `find_node_by_id` (BFS, cycle-guarded) jumps to the target. Dotted `--only a.b` (a node inside a sub-workflow) is rejected with a clear error (membership-checked first), deferred to a name-based follow-up.
- **Trace helpers** (`runtime/workflow_trace.py`), shared by engine **and** planner so they can't drift: `_iter_workflow_traces` (the one candidate iterator), `load_full_run_events`, `load_snapshot_or_raise`, `seed_snapshot_into_shared`. Trace format `2.3.0 → 2.4.0` adds top-level `only_node`, which excludes `--only` runs as snapshot sources — and the same iterator now also keeps `--only` traces from poisoning `analyze-cache` autoload (`prompt_cache_analysis/trace_loading.py`).
- **Display** (`execution/execution_state.py`, `formatters/success_formatter.py`): restored upstream is recorded in `__execution__["restored_nodes"]` and relabeled `not_executed` (fail-closed against `__failures__`), so the execution summary and `--report` agree that only the target ran. The `--only` indicator reads "Ran only 'X'", not "Stopped after".
- **Degraded-source advisory**: a degraded full run can carry partial upstream data (a batch host that continued past failed items). `--only` from such a snapshot is **never silent** — it emits a loud WARNING (`only.snapshot-degraded`) and flips DEGRADED. The degraded check is **severity-aware**: a genuinely degrading WARNING fires it; a benign INFO advisory (empty batch, loop cap) does not.
- **Planner parity** (`execution/plan.py`): `_resolve_walk_start` seeds the snapshot and starts the walk at the target → a single-entry plan (upstream not costed), with the same degraded advisory threaded into `--dry-run`.
- **Robustness**: `OnlySnapshotMissingError` (no usable snapshot → hard error, never a silent re-walk); sub-second trace filenames so a same-second full + `--only` run don't collide.
- **Docs**: rewrote the "re-fires" warning into the snapshot description across `guide/core.md`, `guide/entry.md`, `guide/nodes/llm.md`, `docs/reference/cli/index.mdx`, the Click `--help`, `cli/CLAUDE.md`, and the MCP agent instructions. Added a note that changing inputs needs a full run, not `--only`.

## Explanation

The fix had to satisfy three things that are in tension when an input changes — coherent upstream, no re-fire, and cheap iteration. `--only` deliberately picks "no re-fire + cheap" and accepts "frozen upstream," sourcing it from the **debug trace** rather than the memo cache: the memo cache is write-gated, so it never contains the uncached upstream nodes we need, and `get_latest_for_node` mixes outputs across runs. The trace records every node's output coherently, with no TTL. Rationale (trace vs memo cache vs dedicated store) and the accepted limitations are captured in the ADR.

The design is uniform: `--only` *always* means snapshot, which let the walk's `--only` special-casing be deleted entirely. The engine and planner go through the *same* loader + seed helpers, so `test_plan_drift`-style parity holds by construction.

The change went through two rounds of multi-agent plan review and a round of post-implementation review; the non-obvious findings (the execution-summary regression, the INFO-vs-WARNING degraded distinction, loader de-poisoning, the breaking-test set) are all folded in. Diff: 38 files, ~1.5k insertions.

## Created Docs

- `context/adr/0002-443-only-snapshot-source.md` — why the snapshot source is the debug trace (vs memo cache vs a dedicated store), the trace-coupling trade-off, and the v1 limitations (frozen-upstream + changed-input incoherence, branch freeze, binary/dunder cache hazard, `--no-cache` interaction, degraded-source advisory, same-second collision — resolved).
- `context/CONTEXT.md` — added glossary terms **Snapshot** and **Snapshot vs Cache**.

## Testing

`make test` (7391 passed, 1 skipped) and `make check` (ruff, ruff-format, mypy 227 files, deptry) green. New suite `tests/test_runtime/test_only_snapshot.py` (20 tests) plus updates across 14 existing files. Highlights:

- `test_only_does_not_refire_side_effecting_upstream` — end-to-end: a sentinel-file count proves the side-effecting upstream shell node does **not** re-fire under `--only`, and the summary reports it `not_executed` while only the target is `completed`.
- `test_only_spy_node_upstream_exec_not_called` — engine-level: a restored upstream node's `exec()` is never invoked.
- `test_planner_cached_verdict_matches_engine_serving_cached` / `test_planner_execute_verdict_matches_engine_executing` — real engine-vs-planner verdict-and-behavior parity (not just plan shape).
- `test_real_degraded_batch_snapshot_warns_in_run_and_plan` — a **real** `error_handling: continue` batch (not a synthetic fixture) drives the loud advisory in both run and dry-run.
- `test_degraded_trace_with_info_only_reports_success` — an INFO-only "degraded" trace does **not** false-fire the advisory.
- Loader policy: `--only` traces excluded, `failed` rejected (falls back to older success), empty `nodes` is a no-match, legacy 2.3.0 accepted.

**Manual verification** (real CLI, isolated `$HOME`, adversarial): nested structured restoration (`${produce.result.items[1]}`, `meta.k`) with no re-fire; frozen upstream after editing it; `--no-cache` doesn't defeat the snapshot; branch-divergence (`--only` a node referencing a not-taken branch) → clean loud error listing the snapshot's available nodes, exit 1; no-snapshot → exit 1; loader skips a failed trace and falls back to the older success; 5 rapid repeated `--only` runs all succeed (timestamp-collision fix); declared `## Outputs` referencing a stale downstream node stays target-scoped in display; single-node `--only`, `--report --only` count alignment, and `--dry-run --only` single-entry plan all verified.

## Follow-up

Filed #458 — warn when a changed input feeds a *frozen* upstream node (the documented "incoherent mix" is silent today; an additive detection advisory, out of this PR's scope).


spinje:
# Progress Log — Issue #443: `--only` snapshot semantics

**Branch:** `fix/re-fire-side-effecting-upstream-nodes`
**Status:** Complete. `make test` (7387 non-e2e + 45 e2e) green; `make check` (ruff, ruff-format, mypy 227 files, deptry) green.
**Scope shipped:** FLAT `--only X` snapshot semantics. Dotted `--only a.b` → clear "not supported" error (nested targeting deferred).

---

## 1. The problem (why this exists)

After the cache-defaults flip (`compiler._default_cache_for_node_type` — only `llm` caches by default),
`pflow run wf.pflow.md --only <node>` walked the graph from the start and **re-executed every uncached
upstream node**. Every `shell`/`http`/`mcp`/`file`/`code`/`claude-code` upstream node is a guaranteed
cache miss under the new defaults, so `--only` silently re-fired their side effects on every iteration
(e.g. `shell: gh pr create` creating a PR each run). That defeats `--only`'s purpose ("tune one node
cheaply") and is a real foot-gun.

**Fix:** `--only` no longer walks. It loads the **most-recent full successful run's debug trace**, seeds
the shared store with every node's output except the target, and executes **only the target**. Upstream
is reused, never re-run → no re-fire.

Rationale for trace-as-snapshot-source (vs memo cache vs dedicated store) is in
`context/adr/0002-443-only-snapshot-source.md`. Glossary terms in `context/CONTEXT.md`.

---

## 2. What was built (the shape)

The whole feature funnels through **3 helpers in `runtime/workflow_trace.py`** that BOTH the engine and
the planner call, so the two paths cannot drift:

- `_iter_workflow_traces(debug_dir, workflow_path)` — the single candidate iterator (glob-by-hash, parse,
  collision-guard, `2.x` gate, **exclude `only_node is not None`**). Also now backs the cache-analysis
  autoload (`_collect_candidate_traces`), so an `--only` trace can't poison `analyze-cache` either.
- `load_full_run_events(...)` → `(nodes, status)` — newest reusable run; accepts `success`/`degraded`,
  rejects `failed`, treats empty `nodes` as no-match.
- `load_snapshot_or_raise(...)` — the single home of "no snapshot → `OnlySnapshotMissingError`".
- `seed_snapshot_into_shared(shared, events, *, exclude)` — restore each node's `node_output` to
  `shared[node_id]`, filtering the EXACT `apply_memo_hit` reserved set, never seeding the target.

Engine: `WorkflowEngine.__init__` gained `workflow_path` + `snapshot_events`; `_run_inner` routes flat
`--only` → new `_run_only_snapshot` (seed → run-target-only via `_execute_node` → `_populate_outputs`),
deleting the walk's `--only` branches. `find_node_by_id` (BFS, cycle-guarded) jumps straight to the
target. Planner mirror: `_resolve_walk_start` seeds + starts at target → single-entry plan.

Display: `build_execution_steps` relabels `__execution__["restored_nodes"]` as `not_executed`
(display-only — see §4).

---

## 3. Deviations from the plan (read this section)

### 3a. **The degraded-INFO false-positive — the biggest correctness deviation**
The plan said, literally: *"if `source_status == 'degraded'`: emit a loud advisory."* Implemented naively,
this **broke `test_only_batch_node_runtime_empty_items_emits_compact_summary`**: the `--only` run came
back DEGRADED when the test expected SUCCESS.

Root cause (the key insight of the whole task): **there are two different notions of "degraded".**
- *Result-level* (`runner._is_degrading_warning`) is **severity-based**: INFO advisories don't degrade.
- *Trace-level* (`WorkflowTraceCollector._warning_changes_status`) is **source-based**: any `runtime`
  warning marks the trace `final_status="degraded"`, **regardless of severity**.

An empty-input batch emits a benign **INFO** advisory (per the recent #450/#451 fix "empty-input batch is
a non-degrading advisory"). But that fix only touched the *result* level — the *trace* still records
`degraded`. So keying the loud snapshot advisory off trace `final_status` false-fires on any
INFO-advisory run (empty batch, loop-cap).

**Resolution (deviation):** I added `_trace_has_degrading_warning(data)` and refined `load_full_run_events`
so it returns `status="degraded"` **only when the trace carries a genuinely degrading (WARNING/ERROR)
warning**, mirroring `_is_degrading_warning`. An INFO-only "degraded" trace is reported `success` for
snapshot purposes. This aligns the snapshot's degraded notion with the result-level severity semantics
and keeps the loud advisory honest (the plan's real intent — a batch host with `error_handling: continue`
that *dropped* failed items emits a WARNING, which still fires correctly).

I deliberately did **not** touch `_warning_changes_status` (trace-level status) — that's load-bearing for
analyze-cache trace listing and report rendering; changing it was out of scope and risky.

### 3b. **`TestOnlyNodeWithOutputs` premise inverted (deeper than the plan flagged)**
The plan noted "seeding resolves more declared outputs (intended)". But the existing tests' whole premise
— *"an output sourced from a skipped node is not populated"* — **inverts** under snapshot semantics:
snapshot restores **all non-target nodes, upstream AND downstream**, so `${downstream.x}` now DOES
resolve. The "unresolvable output" case only exists when a source is **absent from the snapshot** (branch
divergence). I reframed all three tests around that, and added a test asserting outputs resolve from a
**restored upstream** node — a capability that didn't exist under the old re-walk-and-stop behaviour.

**Insight for future readers:** snapshot `--only` is NOT "run from start, stop at target." It's "freeze
everything else, run only the target." Downstream nodes are frozen too.

### 3c. **`test_dotted_only_path.py` was a near-total rewrite, not a 3-test patch**
The plan listed updates to 3 tests in this file but flagged "the first review's list was INCOMPLETE."
In reality the file's `TestDottedOnlyIntegration` / `TestDottedOnlyPlanner` / `TestDottedOnlyOutputPopulation`
classes tested the **now-removed** dotted-execution feature (GH #338/#342) end-to-end through the engine —
they all now raise `CompilationError`. I removed them and replaced with `TestDottedOnlyRejected` (cross-entry
`run` vs `plan` rejection parity, membership-first ordering). `TestParseOnlyPath` stayed (the parser is
unchanged).

### 3d. **Timestamp-collision overwrite — a finding NOT in the plan**
Discovered during smoke + the `test_real_cli_only_node_uses_target_not_declared_outputs` test (which runs
`--only` twice). Trace filenames are **second-granular** (`...-YYYYMMDD-HHMMSS.json`). A full run followed
within the same second by an `--only` run writes the **same filename**, so the `--only` trace (which
*excludes itself* as a snapshot source) **overwrites the full-run snapshot** — and the *next* `--only`
then errors with `OnlySnapshotMissingError`.

- Real human iteration is seconds apart → no collision. But fast/scripted/CI runs can hit it.
- Test workaround: `--no-trace` on the `--only` invocations (preserves the seed trace; the snapshot READ
  happens before the WRITE, so a single `--only` run is fine — only repeated same-second `--only`s collide).
- Documented as an ADR limitation. **Not fixed** (filename granularity is pre-existing; out of scope).
  Candidate future fixes: sub-second/uuid suffix in trace filenames, or don't write a trace at all for
  `--only` runs.

### 3e. **`WorkflowRunner().run()` does not persist traces — only the CLI does**
The plan said "write a prior trace." The mechanic the plan didn't state: `WorkflowRunner` builds the
`WorkflowTraceCollector` but **never calls `save_to_file()`** (that's `cli/main.py:_save_trace_and_report`).
So the parity / C1 end-to-end tests that need a disk snapshot must call `full.trace.save_to_file()`
explicitly, marked `@pytest.mark.trace_files`, with `Path.home` patched to the isolated tmp (the autouse
`isolate_pflow_config` already does this at conftest:404).

### 3f. **e2e tests the plan's list omitted**
`tests/test_cli/test_progress_streaming_subprocess.py` has two `--only` subprocess tests (e2e). Both ran
`--only` with no prior run → broke. Fixes:
- `test_print_mode_with_only_emits_indicator`: added a full-run seed subprocess; assertions still hold
  (the "Stopped after 'X' (--only)" check is a substring of the long form).
- `test_only_with_last_node_emits_indicator` → renamed `test_only_with_zero_skipped_emits_indicator` and
  **changed to a single-node workflow**. Why: its intent was "indicator shows even when nodes_skipped == 0"
  (regression guard for an old `if nodes_skipped > 0` gate). Under snapshot semantics, **any multi-node
  `--only` reports nodes_skipped > 0** (the restored nodes count as not_executed), so the only zero-skip
  case is a single-node workflow.

### 3g. **Small mechanical deviations forced by lint/types**
- `find_node_by_id`: plan said "assert returned `.node_id == node_id`". Ruff `S101` forbids `assert` in
  `src/`. Removed it — the `if getattr(node, "node_id", None) == node_id` guard already proves equality.
- Planner seeding: plan put it inline in `_build_plan_with_shared`, but that pushed C901 to 11 (>10).
  Extracted to a helper; the `if/else` still counted as a branch, so I folded the whole branch into
  `_resolve_walk_start(...)` (returns `compiled.start_node` when `this_only is None`). Net function
  complexity back to 10.
- mypy: planner passes `this_only` (narrowed `str`), not the `str | None` `only_node`, to
  `load_snapshot_or_raise`.

### 3h. **Tests scoped out (documented, not written)**
- **Binary-templated-target drift** (specific test): genuinely hard to construct (needs a binary-producing
  node + a memo entry keyed on real bytes vs the sanitized placeholder). The *mechanism* — planner verdict
  matches engine because both seed the same sanitized snapshot — is covered by the two
  engine↔planner parity tests. The binary-specific cache-coherence hazard is documented as an ADR limitation.
- **IR-grew-since-trace** (CLI `nodes_skipped` == `--report` skipped): narrow `--report` edge; skipped for
  time. The general nodes_skipped/restored relabel is covered by the C1 test.

---

## 4. Critical insights & learnings

1. **Two "degraded" notions diverge for INFO runtime advisories** (§3a). This is the single most important
   thing to remember. Trace `final_status` is source-based; result status is severity-based. Anything that
   keys off trace `final_status` to infer "data loss" must add a severity check.
2. **Snapshot restores everything except the target** — upstream AND downstream are frozen. Output
   resolution sees more, not less. "Run from start, stop at target" is the wrong mental model.
3. **Seeding makes `get_node_status` return SUCCEEDED** for restored nodes. This is *correct* for data-flow
   (templates, coalesce, `## Cache` ABSENT→SUCCEEDED rendering should see the restored values). The
   "executed this run" distinction therefore lives **only at the display layer**
   (`build_execution_steps` relabel via `restored_nodes`), never in `get_node_status`. Fail-closed against
   `__failures__` so a real failure can never be hidden by the relabel.
4. **`restored_nodes` changes `nodes_skipped` semantics** — any multi-node `--only` now reports the restored
   nodes as skipped. This is intended and aligns CLI `nodes_executed == 1` with `--report`
   (`nodes_executed == len(events) == 1`).
5. **Trace filenames are second-granular** → same-second full+`--only` overwrite (§3d).
6. **`WorkflowRunner` doesn't save traces; the CLI does** (§3e).
7. The `--only` trace stamps `only_node=<target>`, which **excludes it as a snapshot source** AND (new)
   from analyze-cache autoload — so an `--only` run can never poison either consumer.

---

## 5. Decisions with rationale

| Decision | Chosen | Rationale / alternative rejected |
|---|---|---|
| Default `OnlySnapshotMissingError` suggestion | Baked into the constructor (`_DEFAULT_SUGGESTION`) | DRY — every raise site (engine + planner) gets the good message without repeating the string. Plan showed passing it at the raise site; baking it in is a superset. |
| Degraded-advisory `__warnings__` key | Dedicated synthetic `"__only_snapshot__"`, Diagnostic `node_id=this_only` | A target that emits its own `__warnings__[this_only]` (api_warning / loop cap) would otherwise overwrite the advisory. Dedicated key avoids that; explicit `node_id` keeps the displayed attribution clean (real node, not the synthetic key — `_extract_runtime_warnings` keeps an explicit node_id). |
| `load_full_run_events` empty/failed handling | `continue` (keep scanning older), not `return None` | Strictly more capable: a newer failed/empty trace doesn't shadow an older usable one. Same single-trace result as `return None`. |
| Degraded determination | Severity-aware (`_trace_has_degrading_warning`), not raw `final_status` | §3a — avoids false-firing the loud advisory on benign INFO runs while still firing on genuine WARNING data-loss. Did NOT change trace-level status (out of scope, risky). |
| Dotted `--only` | Hard "not supported" error, membership-checked first | Plan scope. Membership-first so `--only typo.child` still says "not found" with available nodes; `--only realnode.child` says "not supported". Mirrored in engine AND planner so `run`/`plan` reject identically. |
| Shared iterator placement | `runtime/workflow_trace.py`, lazy-imported by `core/prompt_cache_analysis/trace_loading.py` | Keeps the `core ← runtime` edge off import-time (consistent with how trace_loading already lazy-imports runtime). Engine importing `workflow_trace` is a new but light edge (no LiteLLM). |
| Binary-drift & IR-grew tests | Documented as ADR limitation / skipped | §3h — cost/benefit; mechanism is covered by parity + C1 tests. |

---

## 6. Verification

- **Smoke (subprocess, isolated HOME):** full run fires `shell` side effect once (sentinel = 1 line);
  `--only summarize` → output `summary of data`, sentinel **still 1 line** (no re-fire); `--dry-run --only`
  → single-entry plan; fresh `--only` → `OnlySnapshotMissingError`; dotted `--only` → "not supported".
- `make test`: **7387 passed, 1 skipped** (non-e2e). e2e: **45 passed**.
- `make check`: ruff, ruff-format, **mypy 227 files**, deptry — all clean.

New test file: `tests/test_runtime/test_only_snapshot.py` (17 tests): C1 regression (end-to-end + engine
spy), loader source policy (only_node excluded, failed rejected, degraded-WARNING vs degraded-INFO,
2.3.0 accepted, empty no-match), seed unit (reserved keys, target exclude, last-event, no-output skip),
degraded advisory end-to-end, engine↔planner cached/execute verdict parity.

---

## 7. Files changed

**Source:** `core/exceptions.py`, `runtime/workflow_trace.py`, `runtime/engine/engine.py`,
`execution/runner.py`, `execution/plan.py`, `execution/execution_state.py`,
`core/prompt_cache_analysis/trace_loading.py`.

**Docs/CLAUDE.md:** `core/CLAUDE.md`, `runtime/CLAUDE.md`, `runtime/engine/CLAUDE.md`, `execution/CLAUDE.md`,
`cli/CLAUDE.md`, `guide/core.md`, `guide/nodes/llm.md`, `docs/reference/cli/index.mdx`,
`cli/commands/run.py` (Click help), `mcp_server/resources/instructions/mcp-agent-instructions.md`,
`context/adr/0002-443-only-snapshot-source.md` (added Limitations).

**Tests:** new `test_only_snapshot.py`; rewrote `test_dotted_only_path.py`; updated `test_engine_behavior.py`,
`test_cache_integration.py`, `test_workflow_output_handling.py`, `test_cli_mcp_parity.py`, `test_dry_run.py`,
`test_progress_streaming_subprocess.py`, `test_trace_format_2_2.py`, `test_workflow_trace.py`,
`test_cache_analysis_trace_listing.py`; annotated dead-but-green dotted display tests in
`test_output_utils.py` / `test_batch_node.py`; threaded `snapshot_events`/`workflow_path` through
`tests/shared/engine_utils.py` (+ README).

---

## 8. Open items / follow-ups (NOT this PR)

- **Deferred:** name-based, run-anywhere nested targeting (replaces dotted). Build a
  `name → (compiled node, config, trace level)` index, seed that level, run standalone; ambiguous name or
  node-inside-batch-subworkflow → clear "can't isolate" error. The dotted plumbing
  (`_run_node_with_child_only`, `_pflow_child_only_node`, the dotted branches of `find_only_output`) is
  left **dormant** for this — never written today, but kept wired so the follow-up doesn't start from zero.
- **Binary-templated-target drift test** + **IR-grew `--report` consistency test** (§3h) — nice-to-have.
- **Possible future consolidation** (per ADR): merging the debug trace and memo cache into one artifact.
  The snapshot reader would repoint at the merged artifact — building `--only` on the trace today means it
  rides that consolidation naturally rather than being orphaned.

---

## 9. Multi-agent review round (post-implementation)

Ran `/code-review` with 3 specialists (silent-failures, impact-completeness, feature-interactions).
Verdict: no Critical functional bugs; core implementation judged sound and test-pinned. Findings fixed:

- **entry.md doc miss (Critical, impact-completeness).** `src/pflow/guide/entry.md` still taught the OLD
  "the rest re-run" semantics — rendered by both `pflow guide` and `pflow --help` (highest-traffic agent
  surface). I'd updated its 5 siblings but missed this one. Fixed (and tightened: the first rewrite was
  146 chars and wrapped awkwardly in the terse help block).
- **Dry-run planner dropped the degraded status (Warning — silent-failures W1 + impact #2 converged).**
  `_resolve_walk_start` discarded `_source_status`, so `--dry-run --only X` against a degraded snapshot was
  silent while the real run warned — contradicting the ADR's "loud, never silent." Fixed: thread
  `state.diagnostics` into `_resolve_walk_start`; append the same `only.snapshot-degraded` WARNING the
  engine emits (kept inside the helper to stay under C901=10).
- **Three unpinned-but-correct behaviors (feature-interactions W1/W2/W4).** Added regression tests:
  `--only` loop target (one iteration), `--only` top-level WorkflowExecutor target (whole child reruns,
  parent upstream doesn't re-fire, `_pflow_child_only_node` can't leak), and a REAL `error_handling:
  continue` degraded-batch producer→consumer (closes the synthetic-fixture concern, tests/CLAUDE.md #19) —
  the last asserts BOTH the engine advisory AND the new planner diagnostic.
- **`format_only_indicator` wording (impact #3).** `Stopped after 'X' (--only), N remaining nodes skipped`
  → `Ran only 'X' (--only), N other node(s) not executed`. The old text implied a walk-and-stop and that
  nodes were skipped, when only the target ran and the rest were restored — misleading on pflow's
  agent-first surface. Updated the shared formatter + ~15 assertions across 3 test files (grammar tests
  preserved).

Skipped (optional suggestions): `_SNAPSHOT_RESERVED` shared constant (de-risked by trace sanitization),
INFO-producer doc comments, planner `snapshot_events` test-injection parity, MCP `--only` test note.

Verification at this checkpoint (review fixes, before the hardening below): full non-e2e **7390 passed**,
full e2e **43 passed**, `make check` clean. (Final post-hardening counts are at the end of this section.)

### Post-review hardening (approved follow-up)
- **Same-second trace-filename collision (§3d) — FIXED.** `save_to_file` now stamps the filename with
  microsecond granularity (`%Y%m%d-%H%M%S-%f`), so a full run + same-second `--only` run no longer collide
  and the full-run snapshot survives. Safe: the autoload glob keys on the hash prefix; ordering still sorts
  lexicographically; callers passing their own timestamp are unaffected. Pinned by
  `test_save_to_file_filename_is_subsecond_granular`; the `--no-trace` test workaround in
  `test_real_cli_only_node_uses_target_not_declared_outputs` was removed (the fix makes it unnecessary, and
  the test now exercises three real rapid trace writes without collision).
- **Loop-target no-memo-hit** now asserted: `test_only_loop_target_runs_one_iteration` checks
  `__loop_active__` was raised during the run (the scope that suppresses the memo read), not just the
  single-iteration count.

**Final state after hardening:** full non-e2e **7391 passed** (1 skipped), full e2e **43 passed**,
`make check` clean. The de-workaround'd CLI test ran 5× stable (proving the microsecond fix under rapid
in-process trace writes).

### Honest remaining loose ends (acceptable to defer)
- **Documented-but-untested hazards remain:** binary-templated-target cache drift (ADR limitation;
  mechanism covered by the engine↔planner parity tests, no direct test) and IR-grew-since-trace `--report`
  consistency (verified by reasoning + the impact-completeness review, not pinned by a test). Both low-risk;
  flagged for honesty.
- **Trivial:** the planner degraded diagnostic uses `source="planner"` while the engine advisory uses
  `source="runtime"` (same `id`); cosmetic, both render as WARNING.

---

## 10. Independent review + adversarial verification + PR (post-implementation)

A separate reviewer pass (not the implementer) reviewed the staged diff against the plan, then
hammered it through the real CLI trying to break it. Outcome: **faithful to plan, no bugs found.**

### Polish applied (working-tree edits on top of the staged implementation)
- **Dotted-rejection message reworded** in BOTH `_validate_only_target` copies
  (`runtime/engine/engine.py`, `execution/plan.py`): "targeting a node inside a sub-workflow" →
  "targeting a **nested node (dotted path)**"; suggestion "re-run the whole sub-workflow" → "run that
  node alone". The old wording falsely called a plain shell node a sub-workflow when given
  `--only fetch.x` (fetch = shell). Phrase "not supported" preserved, so `match="not supported"`
  assertions still hold.
- **`cli/CLAUDE.md` trace-filename doc** updated to `{...-HHMMSS-ffffff}` with a one-line note on the
  microsecond collision fix.
- **`guide/core.md` `--only` bullet** — added the agent-facing sentence: *"edit the node, not the
  inputs … a changed input reaches only the target; frozen upstream keeps the prior run's value …
  when inputs change, run the full workflow."* (Option 2 from the input-coherence discussion.)

### Adversarial CLI verification (real `.pflow.md`, isolated `$HOME`)
All passed; my own test harness had two bugs (a masked batch failure from a trailing `echo`;
`## Outputs` placed before `## Steps`) which I caught rather than accepting green:
- Nested structured restore (`${produce.result.items[1]}`, `meta.k`) + side-effect NOT re-fired (sentinel).
- Frozen upstream after editing it; `--no-cache` does not defeat the snapshot; editing the target re-runs it fresh.
- Branch divergence (`--only` a node referencing a not-taken branch) → loud clean error listing the
  snapshot's available nodes; exit 1. No-snapshot / `--no-trace` → exit 1.
- **INFO-vs-WARNING degraded distinction verified end-to-end**: empty-batch trace is `final_status=degraded`
  on disk but INFO-only → `--only` treats it as success, NO false advisory. Real `error_handling: continue`
  batch → loud advisory + DEGRADED in run AND dry-run, correct restored `count=3 success=2 err=1`.
- Loader: skips a FAILED trace and falls back to older success; most-recent full run wins; 5 rapid
  repeated `--only` all succeed (timestamp fix); `--only` traces excluded.
- Declared `## Outputs` referencing a stale downstream node → display stays target-scoped (target output,
  not the stale declared value). Start-node `--only` runs the target; single-node → no count;
  `--report --only` counts align (1/2, 1 skipped); `--dry-run --only` → single-entry plan.
- Combining `--only` with a changed input param: works (exit 0), input reaches the target, upstream
  frozen → the documented "incoherent mix" (silent — no warning).

### Follow-up filed
- **#458** — warn when a changed input feeds a *frozen* upstream node. The "incoherent mix" is silent
  today (unlike the degraded-snapshot advisory). Spec'd as an additive detection advisory (reuse the
  per-node `template_resolutions` already in the trace → no format change; precision is automatic
  because only frozen nodes are checked). Explicitly NOT a new "force-cache upstream" execution mode
  (it can't escape the coherence/no-re-fire/cheap trilemma and would reintroduce the #443 re-fire).

### PR
- **#459** — `fix: --only re-runs against a frozen snapshot instead of re-firing upstream (#443)`.
  One commit (38 files); progress log intentionally NOT committed (untracked under `scratchpads/`).
  `make test` 7391 passed / 1 skipped; `make check` clean.

---

## 11. PR #459 review round — three reviews evaluated, fixes landed (commit `f3acdfdc [skip review]`)

Three reviews came in on #459: a strong human-style review, a weak model, and Codex. Each
finding was verified against code (subagent + direct read) before acting — `/evaluate-review`
discipline. Net: **2 disputed-as-false (weak model), 1 disputed-premise (W1), several confirmed.**

### Confirmed & fixed
- **CODEX-1 (real correctness bug, the important one):** `_iter_workflow_traces` reverse-sorted
  the whole filename, which ranks `{safe_name}` ahead of the timestamp. The SAME `workflow_path`
  can have different `safe_name`s (saved-library run vs direct-file run, or a rename), so `--only`
  could restore an OLDER snapshot → stale upstream. Fixed with `_trace_recency_key` (parse the
  trailing `YYYYMMDD-HHMMSS[-ffffff]`, filename as tiebreak). My adversarial pass missed this —
  I always used direct-file runs (constant prefix). Mutation-verified test.
- **CODEX-2 (stale downstream):** `seed_snapshot_into_shared` now seeds only nodes BEFORE the
  target in execution order (`events[:target_idx]`; fallback seed-all if target absent). pflow
  templates only reach earlier steps, so this provably covers every node the target can read while
  excluding downstream — whose stale output was addressable via `-o <downstream>`/`shared_after`.
  Display unchanged (downstream → not_executed either way). Updated 3 tests that encoded the old
  seed-all behavior (`test_only_with_cache` asserted downstream `C` restored; the sub-workflow test
  asserted `post in restored` — both flipped to the correct new contract).
- **CODEX-3 (degraded fail-safe):** a `degraded` trace stays degraded unless its warnings PROVE it
  benign (all INFO/parser-validator); missing/empty warnings → fail-safe degraded, not a silent
  success downgrade. Replaced `_trace_has_degrading_warning` with `_trace_warnings_provably_benign`.
- **W2:** extracted `_workflow_path_id(resolved)` — the 3 duplicated `file_path or synth(ir)` sites
  now call it.
- **S4:** pinned the degraded advisory's `node_id` attribution (the synthetic `__only_snapshot__`
  dict key must not leak as the node id through `_extract_runtime_warnings`).
- **S5:** documented the theoretical cross-process same-microsecond filename collision (no uuid —
  it would break the new timestamp-sort regex; collision is astronomical for sequential agent use).

### Disputed (verified, NOT changed)
- **Weak model — two "critical NameError" claims FALSE:** `Severity` is imported at `plan.py:40`;
  `initialize_execution_state` at `engine.py:48` (the suggested fix even named the wrong module).
  mypy + 7391 tests already proved these.
- **Weak model — `getattr(node, "successors")`:** no real risk; production nodes always have it.
- **S6 (reorder `_iter_workflow_traces` checks "to save a read"):** rationale wrong — the file is
  `json.loads(read_text())`'d before any `.get()` check; reordering saves zero I/O.
- **W1 (lost OutputResolutionError-swallow coverage):** premise wrong — `test_only_node_skips_unresolvable_output`
  still exists (`test_engine_behavior.py:321`) + a CLI test; reviewer mistook an added sibling for a rewrite.
- **S3 (dormant dotted plumbing):** by-design, kept for the deferred name-based follow-up.

### Verification
`make test` **7395 passed / 1 skipped**; `make check` clean. New/updated regression tests
mutation-verified (each fails on the pre-fix code, passes after). Committed as `f3acdfdc` with
`[skip review]`; progress log again intentionally uncommitted.

PR #473, opened 2026-06-05
feat: settable node retry+backoff and status-model consistency (#471)

### LINKED ISSUE #471: Error/resilience model: first-class retry+backoff and a "recovered-cleanly" status (handled failures wrongly mark runs DEGRADED) — audit for more gaps first
## Summary

pflow's error-handling model is incomplete in two ways that force agents to **hand-roll** error logic instead of reusing correct, built-in framework behavior:

1. **No first-class, settable retry-with-backoff for general nodes.** Retry exists but is *hardcoded per node type* and only exposed in `.pflow.md` via `batch:` (per-item). There is no exponential backoff anywhere.
2. **No "recovered-cleanly" run status.** Any `on-error` recovery permanently marks the whole run **DEGRADED** (even when the failure was fully handled), and a retry that eventually succeeds still leaves the run degraded. There is no path from "a failure happened but was handled" back to a clean **SUCCESS**.

This issue tracks fixing the error/resilience model so retry, fallback, and recovery are **built-in, correct, and visualizable** — not re-derived by every agent in `try/except`. **Scope is at least these two problems; a discovery pass will very likely expand it** (saga/compensating rollback, loop×error semantics, batch×error, claude-code soft-fail interaction).

> **Read this as a floor, not a ceiling.** The two headline problems are confirmed; I strongly suspect there are more error/status-model gaps, and Phase 0 below is an explicit audit to find them before implementing.

---

## Why this matters (the principle)

pflow's core bet is that **the framework owns the hard, correctness-critical control flow so agents can't get it wrong.** Error recovery is exactly that kind of thing. Today it's incomplete, so the most natural resilience patterns are either not expressible declaratively or produce a semantically-wrong run status:

- **Retry-with-backoff** — you cannot set retry on a general node in a workflow; you'd have to hand-roll a loop + `try/except` in a code node and compute backoff yourself.
- **Fallback / on-error recovery** — works, but leaves the run flagged DEGRADED even after clean recovery.
- **Compensating rollback (saga)** — appears to have no committed-step tracking and (suspected) also ends DEGRADED after a clean rollback.

The alternative to fixing this — having agents write their own retry/`try/except`/backoff in code nodes — is **fragile and defeats the point**: every agent must independently re-derive correctness, and the framework can't visualize or reason about it. Fix the primitive once.

**Provenance:** this came out of a deep design investigation into pflow's control-flow primitives (loops, branching, error handling). A pattern-coverage analysis of 11 common agent/workflow patterns found the **error-handling family** (retry, saga/rollback) to be the *one* category that is both naturally-expected **and** not cleanly expressible today — and the only one where the gap is a **correctness/semantics** problem, not just ergonomics.

---

## Verified findings (read from current source)

- **`on-error` → DEGRADED, always; no error successor → FAILED.** When a node fails and an `on-error` successor exists, the engine routes to the handler and the run is marked **DEGRADED** (exit 0). With no handler, the run is **FAILED** (exit 1). There is **no** path by which a fully-recovered failure yields a clean SUCCESS.
  - `src/pflow/runtime/engine/engine.py:486` — successor lookup `successors.get(action or "default")`
  - `src/pflow/runtime/engine/engine.py:952-963` — step 17.5 sets `warning=` on `mark_node_failed` when an error handler exists
  - `src/pflow/runtime/engine/engine.py:246-262` — `is_clean_termination(action, successors)`
  - `src/pflow/execution/runner.py` — status determination reads `__warnings__` → DEGRADED (success=True)
- **Three run states only:** `SUCCESS` / `DEGRADED` / `FAILED` — `src/pflow/core/workflow/status.py` (DEGRADED = "completed but some nodes had warnings or non-fatal issues"). No "recovered" state exists.
- **Retry is hardcoded per node type, not settable in workflows.** `Node.__init__(max_retries=1, wait=0)` at `src/pflow/core/node.py:76,84-88`; per-type defaults baked into constructors:
  - http=3, llm=3, file ops=3, claude-code=2, shell=1, code=1
  - A workflow author **cannot** set `max_retries`/`wait` on a general node in `.pflow.md`.
- **Retry is exposed only via `batch:`** — `max_retries` / `retry_wait` in `BATCH_CONFIG_SCHEMA` (`src/pflow/core/ir_schema.py:122-134`), per-item, batched nodes only.
- **No exponential backoff anywhere.** Only a fixed `wait` (float). `grep -rni "backoff\|exponential" src/pflow` returns nothing.
- **The recovered-vs-degraded distinction is already a known sore spot — in the trace layer.** `src/pflow/core/trace_report.py:748-749` ("Old code wrote 'failed' for recovered loops; replace with 'success' since no per-node failures remain") and `:742` ("can't reconstruct the degraded vs success distinction on its own"). So the reporting layer already wrestles with "recovered," but the **run status** still degrades on `on-error`.

---

## Suspected / NOT verified (Phase 0 should confirm or refute each)

Flagging honestly what I have **not** confirmed in code:

- **Saga / compensating rollback specifics** — that there is no committed-step tracking, that compensation ordering is fully manual, and that a clean rollback still ends DEGRADED. Plausible (follows from the verified facts) but reasoned by an independent review agent, **not verified against code**.
- **`on-error` × `loop:` (and backward-edge) interaction** — does a failing node inside a loop let the loop retry? Does it degrade the whole run? Can a loop "recover" across iterations? Unverified.
- **`on-error` × `batch:`** — interaction of per-node retry, batch `error_handling: continue/fail_fast`, and run status. Partially mapped elsewhere, but not for the degraded-status question.
- **Can DEGRADED ever be cleared downstream?** Suspect no (it rides on `__warnings__`). Unverified.
- **claude-code schema soft-fail** routes *outside* `on-error` (verified in earlier design work) — how that interacts with the error/status model is worth folding into the audit.
- **`NonRetriableError` semantics** and exactly which failures are retriable vs not, per node type — only partially mapped.
- **There are almost certainly more gaps** than the two headlined here. **Do not implement before the audit.**

---

## Proposed direction (open for design — NOT final)

Both additions keep error handling declarative, built-in, and visualizable (preserving the strengths agents rely on):

1. **First-class retry/backoff on any node**, e.g.:
   ```yaml
   - retry:
       max: 5
       backoff: exponential   # or fixed
       base: 2s
       on: [transient]        # which error classes retry
   ```
   Generalizes the per-type constructor defaults and the batch-only `max_retries`/`retry_wait` into one settable, consistent param. Open questions: precedence vs existing per-type defaults; batch-level vs node-level; which errors are retriable; how it composes with `on-error`.

2. **A "recovered-cleanly" outcome** so a handled failure (an `on-error` handler that fully recovers, or a retry that eventually succeeds) leaves the run **SUCCESS**, not DEGRADED. Could be a new status, a handler signal (e.g. `recovered: true`), or a status-precedence rule. Must reconcile with the existing trace-layer recovered-loop handling (`trace_report.py`).

Likely in scope **after** the audit: declarative saga/compensation support, and clearer `loop:`×error semantics.

---

## Suggested plan

- **Phase 0 — Audit/discovery (do FIRST).** Verify every item in "Suspected / NOT verified" against code. Map the full state machine: `failure → (retry?) → (on-error?) → run status`, including `loop:`, `batch:`, sub-workflow, and claude-code interactions. **Output: a complete, confirmed gap list + documented current semantics.** Scope for the rest of the issue is set by this.
- **Phase 1 — Status correctness.** Add the recovered-cleanly outcome so handled failures don't falsely degrade the run.
- **Phase 2 — Retry/backoff.** First-class, settable retry+backoff on any node.
- **Phase 3 (likely).** Saga/compensation + loop/error semantics, scoped from Phase 0.

---

## Notes on confidence

- Verified facts above come from reading current source (`engine.py`, `node.py`, `status.py`, `ir_schema.py`, the node constructors, `trace_report.py`).
- The **mechanism** (`on-error` → DEGRADED; retry hardcoded; no backoff) is **verified**. The **judgment** that "degraded-after-clean-recovery is wrong" is a design conclusion (supported by an independent review), not a pre-existing bug report.
- Some supporting analysis (especially the saga specifics) came from a general-purpose review agent on a smaller model and is treated as a **lead to verify**, not ground truth.

spinje:
## Summary

Makes pflow's error/resilience model first-class and internally consistent. Three things, shipped as commits B → C → A:

- **Settable node-level `retry:` with exponential backoff** (was: retry hardcoded per node type, author-settable only via `batch:`).
- **A real `NonRetriableError` short-circuit** (was: a documented-but-no-op marker — deterministic validation failures were retried anyway).
- **A unified run-status model** so on-error recovery, parser/validator advisories, and cached warnings are classified the same way across the runner, trace, CLI, and MCP surfaces (was: two divergent status authorities).

Fixes #471

## Changes

**Settable retry + backoff**
- New top-level `retry:` node field — `max` (1–10, total attempts), `wait` (≥0 finite seconds), `backoff` (`fixed`|`exponential`, clamped at 60s). Applied to the node instance at compile time, so it stays out of the memo cache key and is inherited by parallel-batch deep-copies.
- Inert no-op on `workflow`/non-`Node` nodes (validated, not applied). Strict validation on both the schema and the direct `compile_workflow()` paths, including rejection of non-finite (`.inf`/`.nan`) waits.

**`NonRetriableError` short-circuit**
- Added `PflowError.retriable` (`False` on `CompilationError` and `NonRetriableError`), honored in `Node._exec` and the batch executor (sequential + parallel). Deterministic failures now fail on attempt 1 instead of burning the full retry budget — byte-identical user-facing error/action.

**Unified run-status model**
- One `warning_degrades_status()` predicate (severity **and** source) replaces the runner's severity rule and the trace layer's source-blacklist; both now agree. Fixes a bug where INFO advisories (empty-batch, loop-cap) degraded the trace.
- On-error recovery now emits a structured `Diagnostic` (removes the fragile `— on-error →` string-match), and the trace's `final_status`/`failed_node_ids` treat a recovered node as recovered — previously a recovered run reported trace `failed` while the runner said `degraded`.
- Parser advisories (section typo / unparsed content) reclassified to INFO so a clean run reads `✓ completed` + advisory instead of `⚠️ with warnings`; rendered consistently on CLI run, CLI `--validate-only`, and MCP surfaces.
- The memo cache now round-trips `Diagnostic` warnings (`to_dict`/`from_dict`), so a cached INFO advisory can't flip to degrading on a cache hit (also fixes a pre-existing live LLM-advisory cache bug).

## Explanation

pflow's bet is that the framework owns correctness-critical control flow so agents can't get it wrong. Retry/backoff and recovery-status were the one error-handling gap that was both naturally expected and not cleanly expressible. Rather than add a "recovered" status (the issue's first idea), the change leans on the primitive the codebase already has — `Diagnostic` + `Severity` — and consolidates everything onto one predicate, so "recovered-cleanly" is just SUCCESS-with-an-advisory and the two status authorities can no longer disagree. The on-error → DEGRADED default is deliberately preserved (GH #246); this PR makes the *mechanism* consistent, not the policy. 46 files, +1506/-168.

## Docs
- `docs/how-it-works/batch-processing.mdx`, `src/pflow/guide/core.md`, `src/pflow/guide/features/batch.md` — document `retry:`, the `batch.max_retries × node.max_retries` multiplication (only for re-raising nodes like http/claude), and that `retry:` does not re-run shell non-zero exits (those are `error` actions from `post()`, not raised exceptions).
- `examples/error-handling/retry-with-backoff.pflow.md`, `examples/error-handling/non-retriable-file-error.pflow.md` — runnable examples (auto-validated by `test_docs/test_example_validation.py`).
- `context/CONTEXT.md` — glossary additions: Retry, Backoff, Fallback, Degraded, Advisory.

## Testing
- `test_node_retry.py` — `_retry_delay` fixed vs exponential math + 60s clamp; no sleep after the final attempt; `retriable=False` short-circuits without sleeping; transient errors and `LLMTransientError` still retry.
- `test_retry_config.py` — strict direct-compile validation (rejects float/string/bool/non-finite/unknown-key/out-of-range); `retry:` overrides the per-type default; inert on `workflow`/non-`Node`; parallel-batch workers each inherit the retry budget; editing only `retry:` is a memo cache hit (proves retry is not in the cache key).
- `test_ir_schema_retry.py` + `test_markdown_parser.py::test_non_finite_retry_wait_from_markdown_is_rejected` — schema accept/reject incl. non-finite wait from authored markdown.
- `test_memoization_integration.py` — a `Diagnostic` warning round-trips through the SQLite memo cache (WARNING degrades, INFO does not; severity/message/source intact).
- `test_workflow_trace.py` — a recovered on-error run yields `final_status="degraded"` with `failed_node_ids=[]`; a dict-shaped parser/validator WARNING stays non-degrading.
- `test_on_error_recovery.py` — sub-workflow recovery classified via the propagated structured Diagnostic.
- `test_file_retry.py` — a `NonRetriableError` fails on attempt 1 (replaces the prior test that encoded the no-op bug as expected).
- Manual: `examples/error-handling/retry-with-backoff.pflow.md` (copy-file retried 2×, routed to on-error, completed degraded) and `non-retriable-file-error.pflow.md` (deterministic, no retry sleep); `pflow --validate-only` on a section-typo workflow renders the INFO advisory at exit 0.

## Reviews
Two multi-agent review passes (plan-stage, then staged-code). Notable defects caught and fixed: an undocumented `pool.shutdown(wait=True)` 60s abort-hang regression (reverted to `wait=False`), permissive direct-compile retry coercion + a non-finite `wait` crash hole (hardened on both schema and compile paths), and an on-error-recovery trace/runner status divergence (aligned via per-node recovery tracking).

> Note: the full `make test`/`make check` was not runnable in the dev sandbox (documented `uv`/make panic); verification used the project virtualenv. Recommend a clean `make test` + `make check` in CI before merge.

### DIFF
--- a/context/CONTEXT.md
+++ b/context/CONTEXT.md
@@ -30,6 +30,28 @@ feedback, recurrence, state-threading.
 produced output. From round 2 on, the Carry supplies the value. A role, not a separate field —
 a carried input's ordinary input value *is* its Seed. _Avoid_: initial, default, base.
 
+**Retry** — re-running a *single step's own work* after a **transient** failure, capped by a
+maximum attempt count, same inputs each time. A *deterministic* failure (e.g. bad config) is
+not retried. A retry that eventually succeeds leaves no trace — the run is a clean Success.
+_Avoid_: loop (advances across iterations), fallback, recursion.
+
+**Backoff** — the growing wait between Retry attempts, either *fixed* (constant) or
+*exponential* (doubling), clamped to a ceiling. _Avoid_: delay, sleep, cooldown.
+
+**Fallback (on-error)** — routing to a *different* step when one fails, instead of re-running
+it. The original step genuinely failed, so the run is Degraded (data may be lost) — distinct
+from a Retry that makes the same step succeed. pflow recovery is forward-only: no rollback of
+side effects already done. _Avoid_: catch, rescue, compensation.
+
+**Degraded** — a run that *finished its work but flagged a non-fatal problem* (a Fallback
+fired, a batch dropped failed items, output was salvaged). Completes successfully. Distinct
+from **Failed** (a fatal error halted the run) and from a clean **Success**.
+_Avoid_: partial, warning-state.
+
+**Advisory** — information surfaced about a run that does **not**, on its own, mark it
+Degraded (an empty batch, a Loop hitting its cap, a section typo parsed around). Contrast
+with a degrading warning. _Avoid_: note, hint, info.
+
 **Snapshot** — the frozen prior-run state that `--only <step>` runs a single step against:
 every *other* step's output reused from the most recent full run, so only the target
 re-executes and upstream side effects never re-fire. Requires a prior full run.
@@ -39,6 +61,14 @@ _Avoid_: replay, restore, checkpoint.
 
 **Batch vs Loop** — both repeat a step. Discriminator: can you write the list of runs
 before starting (Batch), or only know you're done by inspecting what just happened (Loop).
+
+**Retry vs Loop** — both re-run a step. Retry re-runs the *same attempt* after a transient
+failure (same inputs, capped, invisible once it succeeds). Loop re-runs across *iterations*,
+each building on the last, until a condition goes falsy.
+
+**Retry vs Fallback** — both are failure responses. Retry re-runs the *same* step hoping it
+succeeds (→ Success). Fallback routes to a *different* step because the original failed
+(→ Degraded).
 
 **Snapshot vs Cache** — both reuse prior output. Cache reuses a step's *own* output when
 its declared inputs are unchanged (correctness-gated, per-step, can still re-run the step).
--- a/docs/how-it-works/batch-processing.mdx
+++ b/docs/how-it-works/batch-processing.mdx
@@ -56,8 +56,8 @@ This runs the `classify` node once for each issue. The `as: "issue"` creates a t
 | `parallel` | bool | No | `false` | Run items concurrently instead of sequentially |
 | `max_concurrent` | int | No | `10` | Maximum parallel items (1-100) |
 | `error_handling` | string | No | `"fail_fast"` | `"fail_fast"` or `"continue"` |
-| `max_retries` | int | No | `0` | Retry failed items this many times |
-| `retry_wait` | int | No | `1` | Seconds to wait between retries |
+| `max_retries` | int | No | `1` | Batch item total attempts after an exception escapes the node (`1` = no batch retry) |
+| `retry_wait` | number | No | `0` | Seconds to wait between batch item attempts |
 
 ## Sequential vs parallel
 
@@ -189,10 +189,31 @@ Process each API call with retries and error tolerance.
     error_handling: continue
 ```
 
-This configuration retries each failed item up to 3 times, waiting 2 seconds between attempts. Common in scenarios involving:
+This configuration gives each failed item up to 3 total batch attempts, waiting 2 seconds between attempts. Common in scenarios involving:
 - Transient API errors
 - Rate limit recovery
 - Network timeouts
+
+Node-level `retry:` is separate from batch retry and applies inside each node attempt:
+
+```markdown
+### fetch
+
+Fetch each API call with exponential node backoff.
+
+- type: http
+- url: ${call.url}
+- retry:
+    max: 3
+    wait: 0.5
+    backoff: exponential
+- batch:
+    items: ${api_calls}
+    as: call
+    parallel: true
+```
+
+For nodes that return an `"error"` action when exhausted (`llm`, `shell`, `mcp`, `code`, file nodes), batch does not run another item attempt. Attempts multiply only for nodes that re-raise after node retries are exhausted, such as `http`, `claude-code`, or custom nodes using the default fallback.
 
 ## What you'll see
 
--- a/examples/error-handling/README.md
+++ b/examples/error-handling/README.md
@@ -18,6 +18,8 @@ These workflows double as regression fixtures for the Task 148 failed-node invar
 | `typo-on-failed-node.pflow.md` | When you typo a field on a failed node (`${primary.stddout}`), the error surfaces BOTH the failure (primary signal) AND the typo correction (secondary hint), and the paste-able fix uses the corrected field |
 | `coalesce-mixed-absent-failed.pflow.md` | A coalesce `${never_run.x ?? fails.y}` with one absent operand and one failed operand errors loudly with the "All coalesce operands are unavailable" summary block — it does NOT silently skip (that was the Task 128 bug class) |
 | `loop-recovery.pflow.md` | A node fails on visit 1, succeeds on visit 2. The loop guard clears the stale `__failures__` entry, the workflow reports success, and downstream references see the visit-2 data |
+| `retry-with-backoff.pflow.md` | A file node uses `retry:` with exponential backoff before exhausting into its `on-error` path |
+| `non-retriable-file-error.pflow.md` | A file-node validation error skips retry sleeps even with `retry:` configured, then routes to `on-error` |
 | `source-line-multi-output.pflow.md` | Multiple output declarations at different lines — the diagnostic's `At:` line points at the exact failing output, not the last output or the top of the file |
 | `source-line-heavy-offsets.pflow.md` | Heavy blank-line padding and prose before the Outputs section — the parser's line tracker handles real-world offsets without off-by-N errors |
 
@@ -29,4 +31,6 @@ These workflows double as regression fixtures for the Task 148 failed-node invar
 pflow examples/error-handling/failed-node-direct-reference.pflow.md --no-cache --no-trace
 pflow examples/error-handling/typo-on-failed-node.pflow.md --no-cache --no-trace
 pflow examples/error-handling/loop-recovery.pflow.md --no-cache --no-trace
+pflow examples/error-handling/retry-with-backoff.pflow.md --no-cache --no-trace
+pflow examples/error-handling/non-retriable-file-error.pflow.md --no-cache --no-trace
 ```
--- a/examples/error-handling/non-retriable-file-error.pflow.md
+++ b/examples/error-handling/non-retriable-file-error.pflow.md
@@ -1,0 +1,52 @@
+# Non-Retriable File Error
+
+Demonstrates that deterministic file validation errors skip the retry wait and route to `on-error`.
+
+## Steps
+
+### prepare-directory
+
+Creates a directory where the copy node expects a file.
+
+- type: shell
+
+```shell command
+mkdir -p /private/tmp/pflow-non-retriable-example
+rm -f /private/tmp/pflow-non-retriable-example/dest.txt
+mkdir -p /private/tmp/pflow-non-retriable-example/source-dir
+```
+
+### copy-directory
+
+Attempts to copy a directory as if it were a file; this is deterministic and should not retry.
+
+- type: copy-file
+- source_path: /private/tmp/pflow-non-retriable-example/source-dir
+- dest_path: /private/tmp/pflow-non-retriable-example/dest.txt
+- retry:
+    max: 3
+    wait: 0.5
+    backoff: exponential
+- on-error: report-error
+- next: done
+
+### report-error
+
+Reports the expected deterministic failure path.
+
+- type: shell
+- next: done
+
+```shell command
+echo "copy-directory failed without retrying deterministic validation"
+```
+
+### done
+
+Marks the end of the recovery path.
+
+- type: shell
+
+```shell command
+echo "non-retriable example complete"
+```
--- a/examples/error-handling/non-retriable-file-error.pflow.md
+++ b/examples/error-handling/non-retriable-file-error.pflow.md
@@ -11,9 +11,9 @@ Creates a directory where the copy node expects a file.
 - type: shell
 
 ```shell command
-mkdir -p /private/tmp/pflow-non-retriable-example
-rm -f /private/tmp/pflow-non-retriable-example/dest.txt
-mkdir -p /private/tmp/pflow-non-retriable-example/source-dir
+mkdir -p /tmp/pflow-non-retriable-example
+rm -f /tmp/pflow-non-retriable-example/dest.txt
+mkdir -p /tmp/pflow-non-retriable-example/source-dir
 ```
 
 ### copy-directory
@@ -21,8 +21,8 @@ mkdir -p /private/tmp/pflow-non-retriable-example/source-dir
 Attempts to copy a directory as if it were a file; this is deterministic and should not retry.
 
 - type: copy-file
-- source_path: /private/tmp/pflow-non-retriable-example/source-dir
-- dest_path: /private/tmp/pflow-non-retriable-example/dest.txt
+- source_path: /tmp/pflow-non-retriable-example/source-dir
+- dest_path: /tmp/pflow-non-retriable-example/dest.txt
 - retry:
     max: 3
     wait: 0.5
--- a/examples/error-handling/retry-with-backoff.pflow.md
+++ b/examples/error-handling/retry-with-backoff.pflow.md
@@ -1,0 +1,52 @@
+# Retry With Backoff
+
+Demonstrates node-level retry with exponential backoff before an `on-error` recovery route.
+
+## Steps
+
+### prepare-path
+
+Ensures the destination path is clear while leaving the source file absent.
+
+- type: shell
+
+```shell command
+mkdir -p /private/tmp/pflow-retry-with-backoff-example
+rm -f /private/tmp/pflow-retry-with-backoff-example/source.txt
+rm -f /private/tmp/pflow-retry-with-backoff-example/dest.txt
+```
+
+### copy-missing-file
+
+Attempts to copy a missing file, exhausts retry attempts, then routes to recovery.
+
+- type: copy-file
+- source_path: /private/tmp/pflow-retry-with-backoff-example/source.txt
+- dest_path: /private/tmp/pflow-retry-with-backoff-example/dest.txt
+- retry:
+    max: 2
+    wait: 0.1
+    backoff: exponential
+- on-error: retry-exhausted
+- next: done
+
+### retry-exhausted
+
+Runs after the copy retry budget is exhausted.
+
+- type: shell
+- next: done
+
+```shell command
+echo "retry budget exhausted" >&2
+```
+
+### done
+
+Marks the end of the recovery path.
+
+- type: shell
+
+```shell command
+echo "retry example complete"
+```
--- a/examples/error-handling/retry-with-backoff.pflow.md
+++ b/examples/error-handling/retry-with-backoff.pflow.md
@@ -11,9 +11,9 @@ Ensures the destination path is clear while leaving the source file absent.
 - type: shell
 
 ```shell command
-mkdir -p /private/tmp/pflow-retry-with-backoff-example
-rm -f /private/tmp/pflow-retry-with-backoff-example/source.txt
-rm -f /private/tmp/pflow-retry-with-backoff-example/dest.txt
+mkdir -p /tmp/pflow-retry-with-backoff-example
+rm -f /tmp/pflow-retry-with-backoff-example/source.txt
+rm -f /tmp/pflow-retry-with-backoff-example/dest.txt
 ```
 
 ### copy-missing-file
@@ -21,8 +21,8 @@ rm -f /private/tmp/pflow-retry-with-backoff-example/dest.txt
 Attempts to copy a missing file, exhausts retry attempts, then routes to recovery.
 
 - type: copy-file
-- source_path: /private/tmp/pflow-retry-with-backoff-example/source.txt
-- dest_path: /private/tmp/pflow-retry-with-backoff-example/dest.txt
+- source_path: /tmp/pflow-retry-with-backoff-example/source.txt
+- dest_path: /tmp/pflow-retry-with-backoff-example/dest.txt
 - retry:
     max: 2
     wait: 0.1
--- a/src/pflow/cli/commands/run.py
+++ b/src/pflow/cli/commands/run.py
@@ -397,6 +397,11 @@ def _display_validation_result(
             if vresult.warnings:
                 for diagnostic in vresult.warnings:
                     click.echo(format_diagnostic(diagnostic), err=True)
+            info_diagnostics = [d for d in vresult.diagnostics if d.severity is Severity.INFO]
+            if info_diagnostics:
+                click.echo("", err=True)
+                for diagnostic in info_diagnostics:
+                    click.echo(format_diagnostic(diagnostic), err=True)
         else:
             from pflow.execution.formatters.validation_formatter import format_validation_failure
 
--- a/src/pflow/cli/commands/run.py
+++ b/src/pflow/cli/commands/run.py
@@ -374,6 +374,20 @@ def _display_execution_result(
         ctx.exit(1)
 
 
+def _echo_diagnostic_group(diagnostics: list[Any], *, blank_line: bool) -> None:
+    """Render a group of diagnostics to stderr, optionally preceded by a blank line.
+
+    Shared by the validation success and failure paths so warnings and INFO
+    advisories are grouped and spaced consistently on both.
+    """
+    if not diagnostics:
+        return
+    if blank_line:
+        click.echo("", err=True)
+    for diagnostic in diagnostics:
+        click.echo(format_diagnostic(diagnostic), err=True)
+
+
 def _display_validation_result(
     ctx: click.Context,
     vresult: Any,
@@ -394,23 +408,16 @@ def _display_validation_result(
             from pflow.execution.formatters.validation_formatter import format_validation_success
 
             click.echo(format_validation_success())
-            if vresult.warnings:
-                for diagnostic in vresult.warnings:
-                    click.echo(format_diagnostic(diagnostic), err=True)
-            info_diagnostics = [d for d in vresult.diagnostics if d.severity is Severity.INFO]
-            if info_diagnostics:
-                click.echo("", err=True)
-                for diagnostic in info_diagnostics:
-                    click.echo(format_diagnostic(diagnostic), err=True)
+            _echo_diagnostic_group(list(vresult.warnings), blank_line=False)
+            _echo_diagnostic_group([d for d in vresult.diagnostics if d.severity is Severity.INFO], blank_line=True)
         else:
             from pflow.execution.formatters.validation_formatter import format_validation_failure
 
             click.echo(format_validation_failure(vresult.errors))
-            extra_diagnostics = [d for d in vresult.diagnostics if d.severity in {Severity.WARNING, Severity.INFO}]
-            if extra_diagnostics:
-                click.echo("", err=True)
-                for diagnostic in extra_diagnostics:
-                    click.echo(format_diagnostic(diagnostic), err=True)
+            # Separate warnings from INFO advisories with a blank line, matching the
+            # success branch above (so advisories read consistently on both paths).
+            _echo_diagnostic_group([d for d in vresult.diagnostics if d.severity is Severity.WARNING], blank_line=True)
+            _echo_diagnostic_group([d for d in vresult.diagnostics if d.severity is Severity.INFO], blank_line=True)
 
     ctx.exit(0 if vresult.valid else 1)
 
--- a/src/pflow/cli/workflow_errors.py
+++ b/src/pflow/cli/workflow_errors.py
@@ -12,6 +12,7 @@ import click
 
 from pflow.core.diagnostic import Diagnostic, Severity
 from pflow.core.diagnostic_render import format_diagnostic
+from pflow.execution.formatters.success_formatter import partition_surfaced_diagnostics
 
 
 def _display_single_error(
@@ -62,11 +63,12 @@ def _display_text_error_details(
     # `result.warnings` property filters to WARNING-only by name; pull from
     # diagnostics directly to include INFO too. Matches the parallel filter
     # in `cli/commands/run.py` on the success path.
-    warnings = [
+    surfaced_diagnostics = [
         diagnostic
         for diagnostic in getattr(result, "diagnostics", [])
         if diagnostic.severity in {Severity.WARNING, Severity.INFO}
     ]
+    warnings, advisories = partition_surfaced_diagnostics(surfaced_diagnostics)
     errors = result.errors
 
     if len(errors) > 1:
@@ -94,3 +96,8 @@ def _display_text_error_details(
         click.echo("⚠️ Warnings:", err=True)
         for warning in warnings:
             click.echo(format_diagnostic(warning), err=True)
+    if advisories:
+        click.echo("", err=True)
+        click.echo("\N{INFORMATION SOURCE}\N{VARIATION SELECTOR-16} Advisories:", err=True)
+        for advisory in advisories:
+            click.echo(format_diagnostic(advisory), err=True)
--- a/src/pflow/core/diagnostic.py
+++ b/src/pflow/core/diagnostic.py
@@ -124,6 +124,21 @@ class Diagnostic:
             result["see_also"] = list(self.see_also)
         return result
 
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> Diagnostic:
+        """Deserialize the structured JSON shape produced by ``to_dict()``."""
+        return cls(
+            severity=Severity(data["severity"]),
+            message=data["message"],
+            title=data.get("title"),
+            suggestions=data.get("suggestions"),
+            node_id=data.get("node_id"),
+            source=data.get("source", ""),
+            context=data.get("context"),
+            see_also=data.get("see_also"),
+            id=data.get("id"),
+        )
+
     def to_display_dict(self) -> dict[str, Any]:
         """Serialize with context merged into top-level keys for display consumers."""
         result = self.to_dict()
@@ -196,6 +211,24 @@ def normalize_runtime_warning(warning: Any) -> tuple[str, dict[str, Any]]:
         return message, context
 
     return str(warning), {}
+
+
+_NON_DEGRADING_SOURCES = frozenset({"parser", "validator"})
+
+
+def warning_degrades_status(entry: Any) -> bool:
+    """Whether a warning entry flips the run/trace status to DEGRADED.
+
+    Shared predicate for live ``__warnings__`` entries and trace
+    ``execution_warnings`` dictionaries. Parser/validator warnings describe
+    definition quality, not runtime data loss, and INFO diagnostics are
+    advisories. Legacy severity-less shapes fail closed as degrading.
+    """
+    if isinstance(entry, Diagnostic):
+        return entry.severity is not Severity.INFO and entry.source not in _NON_DEGRADING_SOURCES
+    if isinstance(entry, dict):
+        return str(entry.get("severity", "")).lower() != "info" and entry.get("source") not in _NON_DEGRADING_SOURCES
+    return True
 
 
 # LLM failure category — single source of truth for the string used by both
--- a/src/pflow/core/exceptions.py
+++ b/src/pflow/core/exceptions.py
@@ -44,6 +44,8 @@ def copy_pflow_annotations(source: BaseException, target: BaseException) -> None
 class PflowError(Exception):
     """Base exception for all pflow errors."""
 
+    retriable: bool = True
+
     def to_diagnostics(self) -> list[Diagnostic]:
         """Convert to diagnostic representation. Override in subclasses for rich output."""
         return [
@@ -781,6 +783,8 @@ class CompilationError(PflowError):
             ``compile_validation.py`` to carry the ``validate_data_flow()`` list
             through the compiler boundary without flattening it to a string.
     """
+
+    retriable = False
 
     def __init__(
         self,
--- a/src/pflow/core/ir_schema.py
+++ b/src/pflow/core/ir_schema.py
@@ -63,6 +63,7 @@ For comprehensive examples, see the examples/ directory.
 """
 
 import json
+import math
 from typing import Any, Union
 
 import jsonschema
@@ -195,6 +196,31 @@ LOOP_CONFIG_SCHEMA: dict[str, Any] = {
 }
 
 
+RETRY_CONFIG_SCHEMA: dict[str, Any] = {
+    "type": "object",
+    "description": "Per-node retry on transient failure (overrides the node-type default). No effect on `workflow` nodes.",
+    "properties": {
+        "max": {
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 10,
+            "description": "Max total attempts (1 = no retry).",
+        },
+        # Keep retry bounds in lockstep with runtime.compilation.compiler's
+        # direct-compile retry validation; the compiler is the safety net for
+        # schema-bypassing dict IR and also rejects non-finite Python floats.
+        "wait": {"type": "number", "minimum": 0, "description": "Base finite seconds between attempts."},
+        "backoff": {
+            "type": "string",
+            "enum": ["fixed", "exponential"],
+            "default": "fixed",
+            "description": "Delay growth mode; exponential waits are clamped at 60s.",
+        },
+    },
+    "additionalProperties": False,
+}
+
+
 # JSON Schema for workflow IR (minimal MVP version)
 FLOW_IR_SCHEMA: dict[str, Any] = {
     "$schema": "http://json-schema.org/draft-07/schema#",
@@ -224,6 +250,7 @@ FLOW_IR_SCHEMA: dict[str, Any] = {
                     },
                     "batch": BATCH_CONFIG_SCHEMA,
                     "loop": LOOP_CONFIG_SCHEMA,
+                    "retry": RETRY_CONFIG_SCHEMA,
                     "_source_lines": {
                         "type": "object",
                         "description": "Markdown source line offsets for code block params (injected by parser)",
@@ -654,6 +681,7 @@ def validate_ir(data: Union[dict[str, Any], str]) -> None:
     if not errors:
         # Additional custom validations
         if isinstance(data, dict):
+            _validate_retry_waits_finite(data)
             _validate_node_references(data)
             _validate_duplicate_node_ids(data)
         return
@@ -669,6 +697,28 @@ def validate_ir(data: Union[dict[str, Any], str]) -> None:
         suggestion=suggestion,
         **context_kwargs,
     )
+
+
+def _validate_retry_waits_finite(data: dict[str, Any]) -> None:
+    """Reject non-finite retry waits that jsonschema's number type admits for Python floats."""
+    nodes = data.get("nodes")
+    if not isinstance(nodes, list):
+        return
+    for index, node in enumerate(nodes):
+        if not isinstance(node, dict):
+            continue
+        retry = node.get("retry")
+        if not isinstance(retry, dict) or "wait" not in retry:
+            continue
+        wait = retry["wait"]
+        if isinstance(wait, bool) or not isinstance(wait, (int, float)):
+            continue
+        if not math.isfinite(float(wait)):
+            raise ValidationError(
+                message=f"{wait!r} is not a finite number",
+                path=f"nodes[{index}].retry.wait",
+                suggestion="Set retry.wait to a finite non-negative number of seconds.",
+            )
 
 
 def _validate_node_references(data: dict[str, Any]) -> None:
--- a/src/pflow/core/markdown_parser.py
+++ b/src/pflow/core/markdown_parser.py
@@ -582,7 +582,7 @@ def parse_markdown(content: str) -> MarkdownParseResult:  # noqa: C901
         else:
             warnings.append(
                 Diagnostic(
-                    severity=Severity.WARNING,
+                    severity=Severity.INFO,
                     message=(
                         f"Unparsed content in '{section_name}' section ({line_ref}). "
                         "Content before the first ### heading is not captured."
@@ -747,7 +747,7 @@ def _resolve_section(section_name: str, line_num: int) -> tuple[_SectionType, bo
     if section_lower in _NEAR_MISS_SECTIONS:
         expected = _NEAR_MISS_SECTIONS[section_lower]
         warning = Diagnostic(
-            severity=Severity.WARNING,
+            severity=Severity.INFO,
             message=f"Line {line_num}: '## {section_name}' looks like a typo for '## {expected}'.",
             suggestions=[f"Rename to '## {expected}'."],
             source="parser",
@@ -1592,12 +1592,12 @@ def _build_node_dict(entity: _Entity) -> tuple[dict[str, Any], dict[str, Any]]:
     # Hoist top-level node fields out of params (they are NOT params). Each must
     # be a top-level key so the params-only unknown-key walk never sees it and the
     # IR-schema check + data-flow carve-outs can inspect ``node[field]`` directly:
-    #   batch / loop — fan-out vs condition-terminated iteration (siblings)
+    #   batch / loop / retry — fan-out, condition-terminated iteration, retry policy
     #   cache        — per-node memoization toggle
     #   prompt_cache / prewarm — Task 159 LLM cache opt-ins; top-level so the
     #     ``cache.invalid-on-non-llm`` rule and IR-schema check (B2.2/B2.3) see them
     #     and validator step 8 doesn't silently allow them on non-LLM nodes.
-    for top_level_field in ("batch", "loop", "cache", "prompt_cache", "prewarm"):
+    for top_level_field in ("batch", "loop", "retry", "cache", "prompt_cache", "prewarm"):
         if top_level_field in all_params:
             node[top_level_field] = all_params.pop(top_level_field)
 
--- a/src/pflow/core/node.py
+++ b/src/pflow/core/node.py
@@ -12,6 +12,8 @@ import time
 import warnings
 from typing import Any, Optional
 
+_MAX_RETRY_BACKOFF_SECONDS: float = 60.0
+
 
 class BaseNode:
     def __init__(self) -> None:
@@ -73,19 +75,30 @@ class Node(BaseNode):
     (2) parallel batch deep-copies the node per thread.
     """
 
-    def __init__(self, max_retries: int = 1, wait: float = 0) -> None:
+    def __init__(self, max_retries: int = 1, wait: float = 0, backoff: str = "fixed") -> None:
         super().__init__()
-        self.max_retries, self.wait = max_retries, wait
+        self.max_retries: int = max_retries
+        self.wait: float = wait
+        self.backoff: str = backoff
 
     def exec_fallback(self, prep_res: Any, exc: Exception) -> Any:
         raise exc
+
+    def _retry_delay(self) -> float:
+        wait = float(self.wait)
+        if self.backoff == "exponential":
+            retry_index = int(getattr(self, "cur_retry", 0))
+            delay: float = wait * (2**retry_index)
+            return min(delay, _MAX_RETRY_BACKOFF_SECONDS)
+        return wait
 
     def _exec(self, prep_res: Any) -> Any:
         for self.cur_retry in range(self.max_retries):  # noqa: B020
             try:
                 return self.exec(prep_res)
             except Exception as e:
-                if self.cur_retry == self.max_retries - 1:
+                if not getattr(e, "retriable", True) or self.cur_retry == self.max_retries - 1:
                     return self.exec_fallback(prep_res, e)
-                if self.wait > 0:
-                    time.sleep(self.wait)
+                delay = self._retry_delay()
+                if delay > 0:
+                    time.sleep(delay)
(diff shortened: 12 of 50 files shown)


spinje:
# Issue #471 — Implementation Plan & Progress Log

This PR was developed plan-first with multi-agent review at each stage: a Phase-0 audit, a review-hardened plan, a staged-code review, and a final evaluation of the two GitHub reviews on this PR. The full implementation plan and progress log (including the review-evaluation pass that produced the latest commit) are included below for reviewers.

<details>
<summary><b>📋 Implementation plan</b> (review-hardened spec — click to expand)</summary>

# Issue #471 — Settable Retry+Backoff & Status Consistency (review-hardened)

**Implementation spec — executable by another agent in isolation.** Every file:line verified against current source (15-agent audit + 4-agent plan review). Code blocks are exact new code; "ANCHOR" names a unique current function/line.

> **v2 — incorporates the 4-agent plan review.** Changed since v1: predicate keeps the source dimension (not severity-only); `NonRetriableError` stays a plain `Exception` with a `retriable` attr (no move to core); C3 adds a local import; A4 guards non-`Node` instances; C6 reconstructs the Diagnostic on rehydration; reclassify ONLY the 2 parser advisories + fix the validate-only renderer; several rationales corrected; missing phases added.

---

## 0. Scope (locked)

**IN — one PR, three commits (B → C → A):**
- **B.** Fix `NonRetriableError` no-op.
- **C.** Status consistency: one shared predicate (`severity != INFO AND source ∉ {parser,validator}`) replacing both status authorities; convert only the on-error recovery writer to a `Diagnostic` (deletes dead reconstruction); fix the cache-boundary Diagnostic round-trip (write+read); reclassify the 2 parser advisories → INFO + render INFO on the validate-only success path.
- **A.** Settable node-level `retry:` + exponential backoff.

**OUT:** saga/compensation; `on-error-clean` flag; 4th status enum; auto-SUCCESS; converting the other legacy `__warnings__` writers (verified unnecessary — they degrade correctly via the predicate's fail-closed branches and round-trip cache-safe).

**Decisions locked:**
- **Predicate keeps the source dimension** (review: severity-only is fragile — an unenumerable set of catalog-based parser/validator WARNINGs would silently degrade clean runs). Status uses source-or-severity; display uses severity alone.
- Reclassify to INFO: **only the 2 parser advisories** (`markdown_parser.py:585`, `:750`). Validator WARNINGs stay WARNING (non-degrading via source dim; visible in validate-only/MCP).
- A1 (retry on sub-workflow nodes): **true no-op** — A4 skips the retry block for non-`Node` instances (NOT a crash). Document-only.
- A2 (retry on mcp/shell/python): allowed; explicit side-effect-guard-override callout in docs.
- A3 `retry.max` = total attempts (1 = no retry). A4 backoff ceiling = 60s. A5 numeric seconds.

---

## Current semantics (verified)
- Run status: `runner._determine_status` (`runner.py:586-605`) reads `__warnings__` (never parser/validator — those go to `__parser_diagnostics__`/`result.diagnostics`). Exit code keys off `result.success` (SUCCESS & DEGRADED both exit 0).
- Two status authorities disagree: `runner._is_degrading_warning` (severity) vs `workflow_trace._warning_changes_status` (source-blacklist).
- Retry: `Node._exec` (`node.py:83-91`), fixed sleep, no backoff, per-type defaults in constructors. Settable only via `batch:`.
- `on-error:` → edge `action="error"`; not a node field.

---

## Commit B — Fix `NonRetriableError` no-op

### B1. `src/pflow/core/exceptions.py`
ANCHOR `class PflowError(Exception):` (line 44). Add a class attribute in the body:
```python
    retriable: bool = True  # False → Node._exec / batch stop retrying immediately
```
ANCHOR `class CompilationError(PflowError):` (line 767). Add `retriable = False` to its class body (deterministic — never retry; aligns with batch's existing "always fatal CompilationError").

### B2. `src/pflow/nodes/file/exceptions.py` — add the marker IN PLACE (no move, no base change)
ANCHOR `class NonRetriableError(Exception):` (line 4). Add ONE attribute; keep the `Exception` base:
```python
class NonRetriableError(Exception):
    """Exception for errors that should not be retried (validation, bad config)."""

    retriable = False
```
**Rationale (corrected from v1):** `Node._exec` checks `getattr(e, "retriable", True)` — it does NOT import or `isinstance`-check the type, so `NonRetriableError` needs no move to core, no `PflowError` base, no re-export change. Keeping the `Exception` base avoids a title regression (a `PflowError` would render via `PflowError.to_diagnostics()` title="Error" instead of the catch-all "Execution Failed") and avoids `except PflowError` newly catching it. Minimal diff.

### B3. `src/pflow/core/node.py` `Node._exec`
ANCHOR `def _exec` (line 83). Add the `retriable` short-circuit (final backoff form is in A3 — **A3 OVERWRITES this body, does not append**):
```python
    def _exec(self, prep_res: Any) -> Any:
        for self.cur_retry in range(self.max_retries):  # noqa: B020
            try:
                return self.exec(prep_res)
            except Exception as e:
                if not getattr(e, "retriable", True) or self.cur_retry == self.max_retries - 1:
                    return self.exec_fallback(prep_res, e)
                if self.wait > 0:
                    time.sleep(self.wait)
```
**MUST** `return self.exec_fallback(...)`, never `raise`/`break`. Output is byte-identical to today's "fail after N retries", just on attempt 1.

### B4. `src/pflow/runtime/engine/batch_executor.py`
ANCHOR `except CompilationError:` inside `_execute_batch_item` (~394). Add a sibling so batch doesn't pointlessly retry a deterministic error:
```python
            except CompilationError:
                raise
            except Exception as e:
                if not getattr(e, "retriable", True):
                    raise  # deterministic — fatal in batch, same as CompilationError (bypasses error_handling:continue)
                last_exception = e
                ...  # existing retry/sleep unchanged
```
**Scope note (corrected):** file-node `NonRetriableError` is raised in `exec()`, consumed by `Node._exec`→`exec_fallback` (returns an `"error"` action), so it never reaches this loop. This branch only fires for exceptions that propagate OUT of `execute_single_fn` — primarily a `python_code` `prep()`-raised `NonRetriableError` (annotation vocab). Aborting the batch for that deterministic error is consistent with `CompilationError`. The comment must say "fatal", NOT "the item's error".

### B5. Tests (Commit B)
- UPDATE `tests/test_nodes/test_file/test_file_retry.py:305-344`: drop the "LESSON LEARNED" comments; assert a **file-node** `NonRetriableError` causes exactly **1** `exec` attempt (spy `exec`; it's raised inside `_validate_source` called from `exec`) yet yields the identical `"Error: ..."` message + `"error"` action. (Do NOT write this for python_code — its `NonRetriableError` is in `prep()`, outside `_exec`.)
- NEW: transient `Exception` still retries to `max_retries`; `LLMTransientError` still retries; a `retriable=False` error short-circuits without sleeping.

---

## Commit C — Status consistency

### C1. `src/pflow/core/diagnostic.py` — single shared predicate + `from_dict`
Add (after `normalize_runtime_warning`):
```python
_NON_DEGRADING_SOURCES = frozenset({"parser", "validator"})

def warning_degrades_status(entry: Any) -> bool:
    """Whether a __warnings__ / execution_warnings entry flips the run to DEGRADED.

    Single source for BOTH the runner (live __warnings__: Diagnostic | dict | str)
    and the trace (execution_warnings: to_display_dict() dicts). Non-degrading iff
    INFO severity OR a parser/validator source (definition-quality, not runtime
    degradation). Severity-less legacy shapes default to degrading (fail-closed).
    """
    if isinstance(entry, Diagnostic):
        return entry.severity is not Severity.INFO and entry.source not in _NON_DEGRADING_SOURCES
    if isinstance(entry, dict):
        return str(entry.get("severity", "")).lower() != "info" and entry.get("source") not in _NON_DEGRADING_SOURCES
    return True
```
Add a `from_dict` classmethod (inverse of `to_dict`, for C6 cache rehydration):
```python
    @classmethod
    def from_dict(cls, data: dict[str, Any]) -> "Diagnostic":
        return cls(
            severity=Severity(data["severity"]),
            message=data["message"],
            title=data.get("title"),
            suggestions=data.get("suggestions"),
            node_id=data.get("node_id"),
            source=data.get("source", ""),
            context=data.get("context"),
            see_also=data.get("see_also"),
            id=data.get("id"),
        )
```

### C2. `src/pflow/execution/runner.py`
**(a)** ANCHOR `_is_degrading_warning` (35-45): body → `return warning_degrades_status(value)`; import the helper; fix the docstring (on-error recovery is now a Diagnostic, not a legacy string).
**(b)** ANCHOR `_extract_runtime_warnings` (613-694): delete the `failures = ...` lookup (616) and the `is_recovery` reconstruction (628-648) — dead once C3 makes on-error recovery a producer-built Diagnostic that short-circuits at line 618. New loop body:
```python
    warnings: list[Diagnostic] = []
    for node_id, raw_message in shared_store.get("__warnings__", {}).items():
        if isinstance(raw_message, Diagnostic):
            warnings.append(raw_message if raw_message.node_id else replace(raw_message, node_id=node_id))
            continue
        message, warning_context = normalize_runtime_warning(raw_message)
        context = {"type": "api_warning"}
        context.update(warning_context)
        if "kind" in warning_context:
            context.setdefault("category", LLM_WARNING_CATEGORY)
        warnings.append(
            Diagnostic(severity=Severity.WARNING, message=message,
                       suggestions=[f"Inspect '{node_id}' upstream inputs and output to verify the warning is expected.",
                                    "If unintended, fix the upstream data or add error handling to this node."],
                       node_id=node_id, source="runtime", context=context))
    # template_errors + parser_diagnostics tails (667-694) unchanged.
```
Update the `_determine_status` docstring (594-596) that names on-error recovery as a legacy string. VERIFIED (silent-failures): on-error recovery is the only producer that set `is_recovery=True`; api_warning/routing are category-excluded today and stay on the api_warning branch.

### C3. `src/pflow/runtime/engine/engine.py` step 17.5 — convert the on-error recovery writer
ANCHOR step 17.5 `if str(action).startswith("error"):` (~952). **Add a local import inside `_execute_node` (Diagnostic/Severity are NOT at module scope — they're function-local elsewhere):**
```python
        from pflow.core.diagnostic import Diagnostic, Severity
        recovery_warning = (
            Diagnostic(
                severity=Severity.WARNING,
                message=f"Node '{config.node_id}' failed — on-error → '{handler_id}'",
                node_id=config.node_id,
                source="runtime",
                context={"type": "on_error_recovery",
                         "category": _NODE_TYPE_FAILURE_CATEGORY.get(config.node_type_name, FAILURE_CATEGORY_NODE_ERROR)},
            ) if handler_id else None
        )
        mark_node_failed(shared, config.node_id,
                         category=_NODE_TYPE_FAILURE_CATEGORY.get(config.node_type_name, FAILURE_CATEGORY_NODE_ERROR),
                         error=node_error, warning=recovery_warning)
```
Sub-workflow recovery now gets a real `context["category"]` (was `None`) — flag any test asserting it is `None`.

### C4. `src/pflow/runtime/node_state.py` `mark_node_failed`
ANCHOR signature (110): widen `warning: Diagnostic | str | None = None` (import `Diagnostic` for the hint). The `__warnings__` write (182) already writes as-is — confirm it does NOT `str()`. KEEP the `str(warning)` at the `__failures__` copy (168) (harmless; its only reader, `runner.py:630`, is deleted in C2b — leave a one-line note).

### C5. `src/pflow/runtime/workflow_trace.py`
- ANCHOR `_warning_changes_status` (835-844): **delete**.
- ANCHOR `_determine_trace_status` (829-830): call `warning_degrades_status(warning)`.
- ANCHOR `_trace_warnings_provably_benign` (143-150): `return all(not warning_degrades_status(w) for w in warnings)`, **keep** the empty/missing-array fail-safe (143-145). (With the source dimension kept, this is behavior-identical to today for parser/validator + fixes the runtime-INFO bug.)
- Import `warning_degrades_status` from `core.diagnostic`.

### C6. `src/pflow/runtime/engine/instrumentation.py` — cache round-trip (write + read)
Import `Diagnostic` at module top (`core/diagnostic.py` is LiteLLM-free).
- ANCHOR `write_memo_cache` (504-506): store the dict form so `json.dumps` keeps severity:
```python
        node_warning = shared.get("__warnings__", {}).get(node_id)
        if node_warning is not None:
            output_dict["__pflow_warnings__"] = node_warning.to_dict() if isinstance(node_warning, Diagnostic) else node_warning
```
- ANCHOR `apply_memo_hit` (397-399): reconstruct the Diagnostic so BOTH `_determine_status` AND `_extract_runtime_warnings` see a real Diagnostic (else the rehydrated dict hits `normalize_runtime_warning`, which keys on `text`/`kind` not `message`/`severity`, corrupting the message to `str(dict)` and forcing WARNING):
```python
        cached_warning = cached_output.get("__pflow_warnings__")
        if cached_warning is not None:
            if isinstance(cached_warning, dict) and "severity" in cached_warning and "message" in cached_warning:
                cached_warning = Diagnostic.from_dict(cached_warning)
            shared.setdefault("__warnings__", {})[node_id] = cached_warning
```
Also fixes a pre-existing live bug: cached LLM INFO advisories (e.g. `cache.routed-provider-degraded`) currently flip to degrading on hit.

### C7. Reclassify the 2 parser advisories `Severity.WARNING` → `Severity.INFO`
- `src/pflow/core/markdown_parser.py:585` (unparsed-section content)
- `src/pflow/core/markdown_parser.py:750` (section near-miss typo)

Leave WARNING (non-degrading via source dim, visible in validate-only/MCP): `data_flow.py:1509` `cache.unused-chunk`, `validator.py:1228` `llm.model-not-in-catalog`, `path_validation.py:193/347` autoparse, and the catalog `cache.prompt-body-shadows-cache`. Both child-propagation paths read the producer Diagnostic, so reclassifying at the producer keeps dedup in sync.

### C7b. Validate-only success path must render INFO advisories
ANCHOR `cli/commands/run.py:397-399` (the `--validate-only` SUCCESS text branch — currently renders only `vresult.warnings`, which is WARNING-only per `execution/result.py:71-73`). Add an INFO-advisory render branch (mirror the run-path `partition_surfaced_diagnostics` advisory section or the validate-only FAILURE branch at ~:404) so the 2 reclassified parser advisories still surface during `pflow --validate-only`. Without this, a section typo silently vanishes from validate-only text output.

### C8. Tests (Commit C)
- UPDATE `tests/test_runtime/test_workflow_trace.py:533`: flip injected `WARNING`→`INFO` (parser), keep `final_status=="success"`.
- UPDATE `tests/test_execution/test_runner.py:800-806` and `:842-847`: the parser near-miss is now INFO — re-point the filters to INFO (preserve the sibling-dedup behavior the `:842` test guards). These break because they assert `severity==WARNING` for the parser producer.
- UPDATE `tests/test_execution/test_on_error_recovery.py` (~198, `test_sub_workflow_recovery_classified_correctly_without_failures`): it injects a raw STRING + exercises the deleted `is_recovery` branch — rewrite to construct a propagated `Diagnostic(context={"type":"on_error_recovery",...})`. Confirm the GH #246 sub-workflow recovery test passes before deleting the dead branch.
- UNAFFECTED (validator diagnostics stay WARNING): `test_workflow_validator_llm_model.py`, `test_template_validation/test_warnings.py`, `test_mcp_warnings.py`, `test_validate_only.py:710` — do NOT need changes (this is the payoff of reclassifying only the 2 parser advisories).
- KEEP `TestDetermineStatusSeverity` (`test_runner.py:1148-1247`) green.
- NEW: cache round-trip — `Diagnostic(WARNING)` and `Diagnostic(INFO)` written to `__warnings__`, cached, rehydrated → reconstructed as Diagnostic with intact severity + message (WARNING degrades, INFO doesn't). (`test_memoization_integration.py` only covers dicts today.)
- NEW e2e: section-typo workflow via `WorkflowRunner().run()` → `status==SUCCESS`, diagnostic `severity==INFO`, trace `final_status=="success"`, CLI `✓`.
- NEW: `pflow --validate-only` on a section-typo workflow → exit 0 AND the advisory text appears in stderr (covers C7b).
- NEW guard: assert the 2 parser advisory producers emit `Severity.INFO`.

---

## Commit A — Settable node-level retry + backoff

### A1. `src/pflow/core/ir_schema.py`
Add after `LOOP_CONFIG_SCHEMA` (~176):
```python
RETRY_CONFIG_SCHEMA: dict[str, Any] = {
    "type": "object",
    "description": "Per-node retry on transient failure (overrides the node-type default). No effect on `workflow` nodes.",
    "properties": {
        "max": {"type": "integer", "minimum": 1, "maximum": 10, "description": "Max total attempts (1 = no retry)."},
        "wait": {"type": "number", "minimum": 0, "description": "Base seconds between attempts."},
        "backoff": {"type": "string", "enum": ["fixed", "exponential"], "default": "fixed", "description": "Delay growth (clamped at 60s)."},
    },
    "additionalProperties": False,
}
```
Add `"retry": RETRY_CONFIG_SCHEMA,` to the node `properties` (after line 208). NOTE (corrected): the node-level `additionalProperties:False` (251) only guards *top-level* keys; a misspelled `reties:` is NOT hoisted (A2), stays in `params` (`additionalProperties:True`, 205), and is caught by validator step-8 unknown-param detection — not a schema error. Do not claim the schema rejects the typo.

### A2. `src/pflow/core/markdown_parser.py:1600`
Add `"retry"` to the hoist tuple: `("batch", "loop", "retry", "cache", "prompt_cache", "prewarm")`.

### A3. `src/pflow/core/node.py` — backoff (OVERWRITES the B3 `_exec` body)
```python
_MAX_RETRY_BACKOFF_SECONDS = 60.0

class Node(BaseNode):
    def __init__(self, max_retries: int = 1, wait: float = 0, backoff: str = "fixed") -> None:
        super().__init__()
        self.max_retries, self.wait, self.backoff = max_retries, wait, backoff

    def _retry_delay(self) -> float:
        if self.backoff == "exponential":
            return min(self.wait * (2 ** self.cur_retry), _MAX_RETRY_BACKOFF_SECONDS)
        return self.wait

    def _exec(self, prep_res: Any) -> Any:
        for self.cur_retry in range(self.max_retries):  # noqa: B020
            try:
                return self.exec(prep_res)
            except Exception as e:
                if not getattr(e, "retriable", True) or self.cur_retry == self.max_retries - 1:
                    return self.exec_fallback(prep_res, e)
                delay = self._retry_delay()
                if delay > 0:
                    time.sleep(delay)
```
`backoff="fixed"` default → every existing node constructor unchanged.

### A4. `src/pflow/runtime/compilation/compiler.py` — apply `retry:`, GUARDED
ANCHOR: immediately AFTER `node_instance.set_params(static_params)` (325). **Guard against non-`Node` instances (`WorkflowExecutor` is a `BaseNode` with no `max_retries` → would `AttributeError`):**
```python
    retry_data = node_data.get("retry")
    if retry_data and isinstance(node_instance, Node):
        node_instance.max_retries = _coerce_retry_int(retry_data.get("max", node_instance.max_retries), "max", node_id, node_type)
        node_instance.wait = _coerce_retry_float(retry_data.get("wait", node_instance.wait), "wait", node_id, node_type)
        node_instance.backoff = retry_data.get("backoff", "fixed")
```
(Import `Node` from `pflow.core.node`.) Runs before batch's `copy.deepcopy(node)` → thread-safe; top-level field → out of the cache key. Add coercers mirroring `_coerce_loop_cap_int` (with `node_id`/`node_type` for attribution):
```python
def _coerce_retry_int(value: Any, field: str, node_id: str, node_type: str) -> int:
    try:
        return int(value)
    except (TypeError, ValueError):
        raise CompilationError(f"Invalid retry config '{field}': expected an integer, got {value!r}",
                               phase="retry_config", node_id=node_id, node_type=node_type)
# _coerce_retry_float analogous (float()).
```

### A5. Edge-case behaviors (verified; corrected claims)
- **Batch × node retry**: distinct loops. Multiplicative `batch.max_retries × node.max_retries` applies **only to nodes whose `exec_fallback` re-raises** (http, claude-code, default-fallback custom nodes). Nodes returning an `"error"` action (llm, shell, mcp, python, file) make `_execute_batch_item` return immediately — **no batch re-retry**. Document precisely; the common LLM target does NOT multiply.
- **Parallel-batch**: SAFE (config set before deepcopy; `time.sleep` releases GIL). fail_fast can't interrupt a worker mid-backoff (up to 60s) — one-line doc note.
- **Loop × retry**: `cur_retry` resets per `_exec`; `clear_node_failure` on re-entry. The real cumulative-latency bound is the loop guard (`MAX_NODE_VISITS`, default 100), NOT the 60s per-attempt clamp — state this accurately.
- **mcp/shell side-effect guard**: `retry:` overrides their deliberate `max_retries=1`; internal retries multiply side effects N×. Explicit docs callout (author asserts idempotency).
- **Sub-workflow node**: A4's `isinstance(node_instance, Node)` guard makes `retry:` a true no-op (no crash). Doc note: child resilience comes from child-internal retries / `on-error`.
- **Cache key**: unaffected (top-level field).

### A6. Tests (Commit A)
- Schema accept/reject (`max` range, `backoff` enum, unknown key).
- `_retry_delay`: fixed=`wait`; exponential=`wait·2^cur_retry` clamped 60; never sleeps after the final attempt; non-retriable skips sleep.
- `retry:` overrides per-type default; omitted preserves it; `retry:{backoff:exponential}` keeps default `max`.
- `retry:` on a `workflow` node compiles cleanly (no AttributeError) and is inert.
- Cache: editing only `retry:` → same cache key (hit).
- Parallel batch: each worker inherits the configured budget.

---

## Cross-cutting invariants (must hold)
1. Short-circuits → `return self.exec_fallback(...)`, never `raise`/`break`.
2. `mark_node_failed` `__warnings__` write (182) must NOT `str()`; the `__failures__` copy (168) stays stringified.
3. `retry:` is a top-level node field, never a `param`.
4. `warning_degrades_status` is the ONLY degrade predicate; it keeps BOTH severity and source dimensions.
5. Cache round-trip needs BOTH `to_dict()` (write) and `from_dict()` (read).
6. A4 retry application is guarded by `isinstance(node_instance, Node)`.
7. C3 step-17.5 needs a local `Diagnostic`/`Severity` import.

## Missing phases (added per review)
- **Agent guide**: update `src/pflow/guide/features/batch.md` (batch×retry multiplication, re-raising-nodes caveat) and `src/pflow/guide/core.md` (the error-handling callout ~line 325) to document `retry:`.
- **Examples**: add `examples/error-handling/` fixtures — a `retry:`-with-backoff workflow and a `NonRetriableError`-no-retry case (auto-validated by `test_docs/test_example_validation.py`).
- **Manual test plan**: run a real `pflow <wf>` with `retry:` + on-error, and `pflow --validate-only <wf>` with a section typo (to confirm C7b renders the advisory and exit 0). These cross-layer paths are exactly what unit tests miss (C6 display, A4 crash, C7b suppression).
- **`/code-review` checkpoint** after Commit C (the riskiest) and again pre-merge.

## Files touched
`core/node.py`, `core/exceptions.py`, `core/diagnostic.py`, `core/ir_schema.py`, `core/markdown_parser.py`; `runtime/compilation/compiler.py`, `runtime/node_state.py`, `runtime/workflow_trace.py`, `runtime/engine/engine.py`, `runtime/engine/batch_executor.py`, `runtime/engine/instrumentation.py`; `nodes/file/exceptions.py`; `execution/runner.py`; `cli/commands/run.py` (C7b); docs (`guide/features/batch.md`, `guide/core.md`, `how-it-works/batch-processing.mdx`), `examples/error-handling/`; tests per B5/C8/A6.

</details>

<details>
<summary><b>📒 Progress log</b> (click to expand)</summary>

# Issue #471 Implementation Progress

## Phase B — Fix `NonRetriableError` no-op

Implemented:
- Added the retry marker contract: `PflowError.retriable = True`, `CompilationError.retriable = False`, and file-node `NonRetriableError.retriable = False` while keeping `NonRetriableError` as a plain `Exception`.
- Updated `Node._exec` to short-circuit `retriable=False` exceptions through `exec_fallback` on the first attempt. This preserves existing node fallback/error-action behavior instead of raising directly.
- Updated batch item retry handling so propagated `retriable=False` exceptions are fatal like `CompilationError`, bypassing batch retry/sleep and `error_handling: continue`.
- Added focused regressions for plain retryable exceptions, `LLMTransientError`, non-retriable no-sleep short-circuit, file-node fast failure with unchanged user-facing error/action, and batch fatal propagation.

Key learnings:
- File-node `NonRetriableError` is raised inside `exec()`, so the core `Node._exec` fast path is the important integration point; changing batch alone would not fix file nodes.
- Batch only sees `retriable=False` when the exception escapes the node execution wrapper, so the batch behavior is intentionally fatal rather than a per-item recoverable error.
- Adding the batch branch pushed `_execute_batch_item` over the existing ruff complexity threshold. I kept a local `# noqa: C901` because refactoring that large function would exceed Phase B scope and risk behavior churn unrelated to the retry marker fix.

Verification:
- `HOME=/private/tmp/pflow-test-home .venv/bin/python -m pytest tests/test_core/test_node_retry.py tests/test_nodes/test_file/test_file_retry.py tests/test_runtime/test_batch_node.py::TestErrorHandling::test_retry_exhausted_exception_error_records_item_summary_and_full_item tests/test_runtime/test_batch_node.py::TestErrorHandling::test_non_retriable_exception_is_fatal_without_batch_retries -q` — 15 passed.
- `HOME=/private/tmp/pflow-test-home .venv/bin/python -m pytest tests/test_runtime/test_batch_node.py -q` — 157 passed.
- `HOME=/private/tmp/pflow-test-home .venv/bin/ruff check src/pflow/core/node.py src/pflow/core/exceptions.py src/pflow/nodes/file/exceptions.py src/pflow/runtime/engine/batch_executor.py tests/test_core/test_node_retry.py tests/test_nodes/test_file/test_file_retry.py tests/test_runtime/test_batch_node.py` — passed.
- `HOME=/private/tmp/pflow-test-home .venv/bin/python -m mypy src/pflow/core/node.py src/pflow/core/exceptions.py src/pflow/nodes/file/exceptions.py src/pflow/runtime/engine/batch_executor.py` — passed.

## Phase C — Status consistency

Implemented:
- Added `Diagnostic.from_dict()` and `warning_degrades_status()` as the shared degradation predicate for runner status, trace status, and degraded snapshot classification.
- Converted on-error recovery warnings to producer-built `Diagnostic` objects and widened `mark_node_failed()` so `__warnings__` preserves structured diagnostics while `__failures__` keeps its legacy string copy.
- Removed runner-side on-error string reconstruction and made memo cache write/read preserve `Diagnostic` warning severity/message/source across SQLite.
- Reclassified only the two parser advisories to `Severity.INFO`, then updated CLI validate-only success output, MCP validation/execution output, and CLI failure output so parser INFO advisories remain visible without being counted as warnings.
- Fixed review-discovered consistency gaps: validation failures now carry non-error diagnostics (not WARNING-only), recovered on-error traces are `degraded` with no unrecovered `failed_node_ids`, and parallel batch `retriable=False` exceptions propagate instead of becoming per-item errors.

Key learnings:
- `ValidationResult.warnings` and `ExecutionResult.warnings` intentionally remain WARNING-only; any surface that must show INFO advisories has to read `diagnostics` directly and partition by severity.
- Child workflow parser diagnostics use fields named `warnings`, but after this phase those lists can contain INFO parser advisories. Renaming that boundary would be larger than Phase C; tests now pin the behavior instead.
- The trace status fix required aligning both `final_status` and `failed_node_ids`; changing only the status would leave trace reports showing recovered nodes as hard errors.
- Two local `# noqa: C901` suppressions remain in batch execution. Both functions were already at the complexity edge; extracting helpers solely to satisfy the threshold would be a phase-local lint refactor, not simpler final behavior.

Review checkpoint:
- Confirmed and fixed: MCP dropped parser INFO advisories on validation/execution surfaces.
- Confirmed and fixed: CLI failure output labeled INFO advisories as warnings.
- Confirmed and fixed: parallel batch swallowed `retriable=False` future exceptions.
- Confirmed and fixed: recovered on-error trace status/failed-node metadata were inconsistent.
- Reviewed but not implemented: adding original node error text to the on-error recovery warning message. The plan specified the concise recovery message; the current change keeps the category and recovery route structured without expanding text output mid-phase.

Verification:
- Focused Phase C tests: 35 passed.
- Review-fix focused tests: 6 passed.
- Broader touched test set: `HOME=/private/tmp/pflow-test-home PATH="$PWD/.venv/bin:$PATH" .venv/bin/python -m pytest tests/test_execution/test_runner.py tests/test_core/test_markdown_parser.py tests/test_runtime/test_only_snapshot.py tests/test_execution/test_on_error_recovery.py tests/test_runtime/test_memoization_integration.py tests/test_runtime/test_node_state.py tests/test_runtime/test_workflow_trace.py tests/test_cli/test_validate_only.py tests/test_runtime/test_batch_node.py tests/test_mcp_server/test_validation_service.py tests/test_mcp_server/test_execution_workflow.py tests/test_cli/test_direct_execution_helpers.py -q` — 623 passed.
- `ruff check` on all touched Phase C/B files — passed.
- `mypy` on touched production files — passed.
- Manual: `HOME=/private/tmp/pflow-test-home PATH="$PWD/.venv/bin:$PATH" .venv/bin/pflow scratchpads/issue-471-retry-status/manual/section-typo.pflow.md --validate-only` — exit 0, rendered the parser INFO advisory.

## Phase A — Settable node-level retry + backoff

Implemented:
- Added top-level `retry:` IR schema support, markdown hoisting, compiler application, and guarded handling for non-`Node` instances.
- Extended `Node` with `backoff`, `_retry_delay()`, a 60s exponential clamp, and typed retry state while preserving the existing `exec_fallback()` short-circuit contract for `retriable=False`.
- Added compiler retry coercion for direct `compile_workflow()` callers, including bounds and backoff validation so invalid in-memory IR cannot compile into a non-executing node.
- Added retry docs in the guide and public batch docs, plus runnable error-handling examples for retry exhaustion and deterministic non-retry recovery.
- Added focused schema, parser, node retry, compiler, cache-hash, non-`Node`, and parallel-batch retry tests.

Key learnings:
- `retry:` belongs on the node instance, not `NodeConfig`; this keeps retry changes out of memo cache keys by construction.
- Shell non-zero exits are action failures from `post()`, not exceptions from `exec()`, so they do not exercise `Node._exec` retry. HTTP connection failures do retry but `HttpNode.exec_fallback()` raises a terminal `ValueError`, so it cannot demonstrate `on-error` routing. The manual retry example uses `copy-file` with a missing source because that path retries an `exec()` exception, then returns an `"error"` action that `on-error` can route.
- The plan only specified simple retry coercers; I added direct-compile validation for malformed type/range/backoff because the CLI/schema path is not the only caller of `compile_workflow()`, and `retry.max: 0` would otherwise create a node that never executes.

Deferred with rationale:
- Did not run the pre-merge `/code-review` checkpoint in this phase. Current instruction is to stop after Phase A and wait for human review; the pre-merge checkpoint belongs after the human accepts/stages this final phase, not before this requested review boundary.

Verification:
- `HOME=/private/tmp/pflow-test-home .venv/bin/python -m pytest tests/test_core/test_node_retry.py tests/test_core/test_ir_schema_retry.py tests/test_core/test_markdown_parser.py::TestSectionHandling::test_retry_field_is_hoisted_to_node_top_level tests/test_runtime/test_retry_config.py tests/test_docs/test_example_validation.py -q` — 24 passed.
- `HOME=/private/tmp/pflow-test-home .venv/bin/ruff check src/pflow/core/ir_schema.py src/pflow/core/markdown_parser.py src/pflow/core/node.py src/pflow/runtime/compilation/compiler.py tests/test_core/test_node_retry.py tests/test_core/test_ir_schema_retry.py tests/test_core/test_markdown_parser.py tests/test_runtime/test_retry_config.py` — passed.
- `HOME=/private/tmp/pflow-test-home .venv/bin/python -m mypy src/pflow/core/ir_schema.py src/pflow/core/markdown_parser.py src/pflow/core/node.py src/pflow/runtime/compilation/compiler.py` — passed.
- Manual: `HOME=/private/tmp/pflow-test-home PATH="$PWD/.venv/bin:$PATH" .venv/bin/pflow examples/error-handling/retry-with-backoff.pflow.md --no-cache --no-trace` — exit 0, copy-file retried twice, routed to `retry-exhausted`, and completed degraded.
- Manual: `HOME=/private/tmp/pflow-test-home PATH="$PWD/.venv/bin:$PATH" .venv/bin/pflow examples/error-handling/non-retriable-file-error.pflow.md --no-cache --no-trace` — exit 0, deterministic directory-as-file error routed to `report-error`.
- Manual: `HOME=/private/tmp/pflow-test-home PATH="$PWD/.venv/bin:$PATH" .venv/bin/pflow scratchpads/issue-471-retry-status/manual/section-typo.pflow.md --validate-only` — exit 0, rendered the parser INFO advisory.

## Phase A — Adversarial manual verification pass

Implemented during verification:
- Added scratchpad manual workflows covering retry exhaustion with `on-error`, non-retriable fast path, shell post-error no-retry boundary, invalid retry schema, workflow-node retry no-op, and retry-only memo cache reuse.
- Fixed a real regression found by the broad suite: `Node.__init__` briefly initialized `cur_retry`, which changed direct `exec()` behavior for nodes that intentionally test `hasattr(self, "cur_retry")`. Restored the prior boundary by only setting `cur_retry` during `_exec()` and making `_retry_delay()` tolerate absence.
- Fixed a stale Phase C test expectation in `WorkflowExecutor` child parser diagnostic propagation from WARNING to INFO.
- Fixed misleading file-node fallback logs so deterministic `NonRetriableError` paths no longer print "after N retries" when they correctly short-circuit.
- Tightened guide wording after manual probing showed `retry:` does not re-run shell non-zero exits because those are `"error"` actions from `post()`, not raised `exec()` exceptions.

Key verification findings:
- `copy-file` missing source with `retry.max: 3`, `wait: 0.05`, `backoff: exponential` produced three source-missing attempts, took ~0.20s, then routed to `on-error`.
- `copy-file` directory-as-file with `retry.max: 3`, `wait: 0.5` produced one deterministic validation failure and completed in ~0.04s, proving the non-retriable fast path skips retry sleep.
- Shell non-zero exit with `retry.max: 3` ran exactly once (`attempts` output `1`), confirming the post-action boundary.
- Changing only `retry.max` on a `cache: true` shell workflow reused the cached output (`1`) and left the marker file at `1`, confirming retry config is not in the memo cache key.
- `retry:` on a `workflow` node validated and executed successfully as an inert field; child output propagated.

Verification:
- Read via actual CLI: `pflow --help`, `pflow guide core`, `pflow guide batch`, `pflow guide branching`, and `pflow guide file shell http sub-workflows`.
- Manual validation: all positive scratchpad workflows passed `--validate-only`; `phase-a-invalid-retry-backoff.pflow.md` failed with `'linear' is not one of ['fixed', 'exponential']` at `nodes[0].retry.backoff`.
- Manual execution: `phase-a-retry-copy-missing.pflow.md`, `phase-a-nonretriable-fast-path.pflow.md`, `phase-a-shell-post-error-no-retry.pflow.md`, `phase-a-workflow-retry-noop-parent.pflow.md`, and `phase-a-cache-retry-toggle.pflow.md` produced the expected behavior above.
- Broad sandbox suite: `HOME=/private/tmp/pflow-test-home .venv/bin/python -m pytest -n 4 --doctest-modules --ignore=tests/test_nodes/test_llm/test_llm_integration.py -k 'not test_dry_run_json_mode_emits_no_stderr and not test_litellm_not_imported_by_cli_main and not test_progress_streams_before_downstream_nodes_complete'` — 7587 passed, 19 skipped, 4 failed from the known `/opt/homebrew/bin/uv` sandbox panic class.
- Targeted real-regression rerun: `tests/test_runtime/test_flow_construction.py::TestInstantiateNodes::test_instantiate_multiple_nodes_creates_different_instances tests/test_runtime/test_workflow_executor/test_workflow_executor.py::TestWorkflowExecutor::test_prep_preserves_child_parser_warnings_when_input_validation_fails tests/test_core/test_node_retry.py` — 10 passed.
- File retry/log regression rerun: `tests/test_nodes/test_file/test_file_retry.py tests/test_core/test_node_retry.py` — 18 passed.
- `mypy` on touched production files — passed.
- `ruff check` on touched production/test files — passed.
- `git diff --check` — passed.

## Post-review fix pass — Retry validation and review hardening

Implemented:
- Reverted the off-plan parallel batch shutdown change back to `pool.shutdown(wait=False, cancel_futures=True)` to preserve fail-fast abort latency under node backoff sleeps.
- Hardened direct-compile `retry:` validation: unknown keys are rejected, `max` must be a real integer in range, `wait` must be a real finite non-negative number, `backoff` must match the enum, and the same field validation runs for inert non-`Node` / `workflow` nodes before optional application.
- Added schema-side finite validation for `retry.wait` so YAML `.inf` / `.nan` from authored markdown fail before execution; clarified docs that only exponential waits are 60s-clamped while fixed waits use the declared finite value.
- Ratified the review-era intent expansions instead of reverting them: validator non-error diagnostics remain surfaced, MCP parser/validator INFO diagnostics remain visible as `advisories`, and recovered on-error trace status stays aligned through `_unrecovered_failed_node_ids`.
- Removed the dead stringified `__failures__[node_id]["warning"]` mirror; structured warnings now live only in `__warnings__`, which is the actual consumer boundary.
- Added regression coverage for strict direct compile retry validation, invalid retry on inert/workflow nodes, real memo-cache reuse when only retry config changes, markdown/schema non-finite waits, dict-shaped parser/validator WARNING trace status and `--only` snapshot classification, and on-error recovery category preservation.

Key learnings:
- JSON Schema's `number` check accepts Python `float("inf")` and `float("nan")`, so finite numeric constraints need an explicit custom validation pass.
- "Inert on workflow/non-Node nodes" must mean "validated but not applied"; otherwise direct `compile_workflow()` becomes a schema bypass for invalid config.
- Retry config is correctly outside the memo cache key, but a direct hash-only test was too synthetic; the production signal is a second run with no repeated side effect and restored cached output.
- The old failure-record warning mirror had become misleading after recovery warnings moved to producer-built `Diagnostic` objects. Keeping only `__warnings__` preserves the structured consumer contract and removes stale duplicate state.

Deviations from plan:
- Implemented the optional `__failures__.warning` cleanup now instead of deferring. Rationale: search found no production readers and the existing test only encoded the stale mirror, so removing it reduced ambiguity without expanding behavior.
- Did not run `make test` / `make check` directly because this sandbox's project guidance documents `uv`/make panics. Used the virtualenv commands from `pflow-sandbox-testing` instead.

Verification:
- Focused retry/schema/status set: `HOME=/private/tmp/pflow-test-home .venv/bin/python -m pytest tests/test_runtime/test_retry_config.py tests/test_core/test_ir_schema_retry.py tests/test_core/test_markdown_parser.py::TestSectionHandling::test_non_finite_retry_wait_from_markdown_is_rejected tests/test_runtime/test_workflow_trace.py::TestWorkflowTraceCollector::test_final_status_success_with_definition_warning_dict tests/test_runtime/test_only_snapshot.py::test_degraded_trace_with_definition_warning_reports_success tests/test_execution/test_on_error_recovery.py::test_sub_workflow_recovery_classified_correctly_without_failures tests/test_runtime/test_node_state.py::TestClearNodeFailure::test_preserves_diagnostic_warning_only_in_warnings_channel -q` — 36 passed.
- Broader touched suites: `HOME=/private/tmp/pflow-test-home PATH="$PWD/.venv/bin:$PATH" .venv/bin/python -m pytest tests/test_runtime/test_retry_config.py tests/test_core/test_ir_schema_retry.py tests/test_core/test_markdown_parser.py tests/test_runtime/test_batch_node.py tests/test_runtime/test_workflow_trace.py tests/test_runtime/test_only_snapshot.py tests/test_execution/test_on_error_recovery.py tests/test_runtime/test_node_state.py tests/test_runtime/test_memoization_integration.py tests/test_cli/test_validate_only.py tests/test_mcp_server/test_validation_service.py tests/test_mcp_server/test_execution_workflow.py -q` — 576 passed.
- Final focused retry/schema rerun: `HOME=/private/tmp/pflow-test-home .venv/bin/python -m pytest tests/test_runtime/test_retry_config.py tests/test_core/test_ir_schema_retry.py tests/test_core/test_markdown_parser.py::TestSectionHandling::test_non_finite_retry_wait_from_markdown_is_rejected -q` — 30 passed.
- `HOME=/private/tmp/pflow-test-home .venv/bin/ruff check ...` on touched production/test files — passed.
- `HOME=/private/tmp/pflow-test-home .venv/bin/python -m mypy src/pflow/runtime/compilation/compiler.py src/pflow/core/ir_schema.py src/pflow/runtime/engine/batch_executor.py src/pflow/execution/runner.py src/pflow/runtime/node_state.py` — passed.
- `git diff --check` — passed.

## Staged-code review (second review pass) + finalization

Reviewed the staged implementation with 5 parallel review agents (silent-failures,
impact-completeness, test-fidelity, feature-interactions, concurrency-safety) plus a
manual read of every critical production hunk. Findings consolidated in
`staged-code-review-synthesis.md` (merged with a prior validation-focused pass).

Verdict: faithful, high-quality implementation that sensibly gap-filled four
under-specified spots in the plan. Findings → all addressed (see the "Post-review fix
pass" section above):
- P0 `pool.shutdown(wait=True)` 60s abort-hang regression → reverted to `wait=False`.
- P0 permissive direct-compile retry coercion + non-finite `wait` crash hole → hardened
  on both schema and compile paths; field validation now runs before the non-`Node` guard.
- P1 scope expansions (validator non-error diagnostics, MCP parser/validator INFO
  advisories, on-error recovery trace alignment via `_unrecovered_failed_node_ids`) →
  ratified as correct, not reverted.
- P2 cleanups → stale docs updated, lockstep comment added, dead `__failures__.warning`
  mirror removed.

Review-process notes (artifacts, not defects):
- Two "alarms" during the parallel review were artifacts of review agents concurrently
  running `git stash`/pytest in the same worktree — a false "working tree reverted"
  report (misread of staged-but-uncommitted state) and an apparent test flake (the
  flagged 5-file combo is deterministically green, 182 passed × 3 clean single-process
  runs). The staged index was verified intact throughout. Lesson: do not run multiple
  tree-mutating agents concurrently against one worktree.

Final doc-drift fix (this session): the `__failures__.warning` removal left two stale
record-shape docs describing a field that is no longer written — fixed
`node_state.py::get_node_failure` docstring and `runtime/CLAUDE.md` `__failures__`
reference (pure doc edits, no behavior change).

## PR

- Branch `feat/retry-backoff-status-471` (renamed from the placeholder
  `fix/resolve-invalid-config-path`, fast-forwarded onto `main`).
- Commit `aba70af6` — `feat: settable node retry+backoff and status-model consistency
  (#471)`; 46 files, +1506/-168; `scratchpads/` intentionally excluded; pre-commit hooks
  passed.
- PR: https://github.com/spinje/pflow/pull/473 (base `main`, `Fixes #471`).
- Pre-merge gate still open: a clean `make test` + `make check` in CI (the dev sandbox
  could not run `make` due to the documented `uv`/make panic; verification used the
  project virtualenv throughout).
- CI on `aba70af6`: green — `tests-and-type-check` passed on Python 3.10–3.14, `quality`
  (lint+types) passed, automated `claude-review` passed. The 4 local "failures" were
  confirmed sandbox artifacts.

## Review-evaluation + fix pass (commit `c3459ac9`)

Evaluated the two GitHub PR reviews against actual code (6 parallel
`pflow-codebase-searcher` agents — 3 for the review findings, 3 for an independent
architect reconciliation of the implementation vs the plan). **Do-not-trust-the-review**
discipline: both reviewers' flagged-serious findings turned out to be disputed.

Scorecard:
- **claude#1** (`error_handling: continue` aborts on a per-item non-retriable error):
  **DISPUTED**. File-node `NonRetriableError` is converted to an `"error"` *action* inside
  `Node._exec`→`exec_fallback`→`post`, so it never escapes as an exception into the batch
  `except`/`raise`. Only definition-style errors that genuinely escape `node._run`
  (python-code `prep()` annotation errors, `CompilationError`) abort — identical across all
  items, so aborting is correct. No regression.
- **G1** (`_unrecovered_failed_node_ids` misses `Diagnostic` objects): **DISPUTED**.
  `execution_warnings` is always dicts (`set_warnings`→`to_display_dict`, which merges
  `type`/`node_id` to top-level); field is typed `list[dict]`. Existing test already pins
  recovered nodes → `failed_node_ids == []`.
- **G2** (cancel pending futures before raise): **DISPUTED / already-handled**. Bare
  executor + `finally: pool.shutdown(wait=False, cancel_futures=True)` (Task 131 decision);
  the proposed fix would be redundant. Also confirms the synthesis's P0/A `wait=True`
  regression was reverted in the commit.
- **claude#5** (repeated local import): non-issue — follows the existing deferred-import
  pattern (reviewer agreed).
- **claude#2** (`retry:` on a `workflow` node validated but silently inert): **FIXED** — the
  validator now emits an INFO advisory (`_retry_on_workflow_node_advisory`,
  `source="validator"` → non-degrading) via the existing per-type dispatch.
- **claude#3** (advisory presentation inconsistent across error surfaces): **FIXED on both
  surfaces** — the CLI (`run.py`) and MCP (`execution_service.py`) validation-FAILURE
  branches now separate WARNING from INFO advisories like their success branches (CLI via a
  new `_echo_diagnostic_group` helper that also dropped the function back under the C901
  threshold; MCP by reusing `_mcp_surfaced_diagnostics` + `_format_mcp_diagnostic_sections`).
- **claude-nit** (example `/private/tmp` paths): **FIXED** — `/tmp` in the two
  `examples/error-handling/` fixtures (Linux-portable; CI unaffected — validate-only).
- **claude#4** (broaden `retriable=False` to other deterministic `PflowError`s): noted as a
  follow-up, not this PR.
- **MCP `advisories` JSON-key** (synthesis P1/D): **ratified** (already sound + documented +
  tested) and made fully consistent via the claude#3 MCP fix.
- **synthesis gap L** (source-dimension not pinned): **closed** — added a direct unit test on
  `warning_degrades_status` so a regression to a severity-only predicate fails loudly.

Architect reconciliation: all six status invariants and all retry invariants **matched the
plan**; the `__failures__.warning` removal is clean (zero readers, docs updated).

Changes (8 files, +154/-26): `run.py`, `validator.py`, `execution_service.py`, 2×
`examples/error-handling/*.pflow.md`, `test_diagnostic.py`, `test_validate_only.py`,
`test_validation_service.py`.

Verification (project venv): all new tests pass; validate-only / validator / diagnostic /
MCP-validation / MCP-execution / example-validation / failed-node-invariant suites pass;
`ruff` + `mypy` clean on every touched file. Commit `c3459ac9` (`[skip review]`) pushed;
pre-commit hooks passed.

</details>

Issue #494, opened 2026-06-07
Use abbreviated token counts (2.1M, 153K) consistently across human-facing displays


spinje:
## Summary

Token counts are rendered inconsistently long across human-facing surfaces — e.g. a cache-heavy agent run shows `Tokens: 2,137,122 in / 41,733 out` and `💰 Cost: $0.05 (9 calls, 2,137,122 tokens)`. We want **abbreviated** counts (`2.1M`, `153K`) for human-facing displays, applied **consistently** through one shared helper so no surface drifts.

Motivation: this came out of the #492 follow-up. We considered abbreviating the cost line alone but rejected it — abbreviating one surface while others use `:,` creates cross-surface inconsistency. The right fix is to abbreviate **everywhere a human reads a token count**, via a single seam.

## Proposed solution

Add one shared helper (mirroring `core/duration_format.format_duration`), e.g. `core/token_format.format_token_count(n: int) -> str`, and route every human-facing token-**count** display through it.

Proposed rule (this is the one real **decision** — tweak as desired):

| Range | Render | Example |
|---|---|---|
| `n < 10_000` | exact, comma-grouped | `1,234` |
| `10_000 ≤ n < 1_000_000` | `K`, 0–1 decimals | `153K` |
| `n ≥ 1_000_000` | `M`, 1 decimal | `2.1M` |

Small counts stay exact (abbreviation buys nothing under ~10K and hurts precision). Keep the cache-% etc. computed from the raw ints — only the displayed string changes.

## In scope — human-facing token COUNTS

- `core/trace_report.py`
  - `_format_tokens` → `{total_in:,} in / {tokens_out:,} out` (summary line, per-node line, pipeline table cell)
  - thinking-tokens line (`~:413`)
  - `Cache write:` / `Cache read:` lines (`~:1249`, `~:1251`)
- `cli/workflow_output.py` — CLI cost line `{tokens_total:,} tokens` (`~:685`)
- `execution/formatters/success_formatter.py` — MCP cost line `{tokens_total:,} tokens` (`~:368`)
- `core/prompt_cache_analysis/` text renderers — token **counts** only:
  - `rendering/text.py` — cache-read/write token counts, "historical input tokens", `~{tokens:,}` estimates
  - `rendering/cross_workflow_edits.py::_format_tokens_phrase` (`~{tokens:,} tokens`)
  - `stages/partial_declarations.py` — the `~{tokens:,} tokens` count portion

[...]

Issue #18, opened 2025-12-29
Template variables: arrays cannot be embedded in shell command strings


spinje:
## Problem

When a template variable resolves to an array or dict, it cannot be used in shell command strings. The type checker rejects it at compile-time:

```
Type mismatch in node 'process-array' parameter 'command':
template ${items} has type 'array' but parameter expects 'str'
```

## Reproduction

```json
{
  "inputs": {
    "items": {
      "type": "array",
      "required": true
    }
  },
  "nodes": [
    {
      "id": "process-array",
      "type": "shell",
      "params": {
        "command": "echo '${items}' | jq 'length'"
      }
    }
  ],
  "edges": [],
  "outputs": {}
}
```

**Expected**: Array serializes to JSON and the command works
**Actual**: Validation error before execution

## Current Workaround

Use `stdin` parameter instead:

```json
{
  "stdin": "${items}",
  "command": "jq 'length'"
}
```

**Limitation**: Only ONE value can be passed via `stdin`. If you need multiple arrays in one command, there's no good solution.

## Real-World Use Case

Building a workflow that:
1. Downloads images → outputs array of paths
2. Describes images with vision LLM (batch) → outputs array of descriptions
3. Needs to combine BOTH arrays in a shell command to build a mapping

Step 3 is currently impossible because we can only pass one array via stdin.

## Proposed Solution

1. **Allow array/dict → str in type checker** - When an array or dict appears in a string context, auto-serialize to JSON
2. **Keep runtime check for simple templates** - `command: "${items}"` (just the variable) should still fail, but `command: "echo '${items}'"` (embedded in text) should work
3. **Add warning for shell-unsafe JSON** - If serialized JSON contains apostrophes, backticks, or `$()`, warn user to use stdin instead

## Considerations

- Shell escaping is complex - JSON with apostrophes (`it's`) breaks single-quoted shell strings
- This should work for most cases but isn't foolproof
- Document limitation: "For data with special characters, use stdin"

PR #495, opened 2026-06-07
fix: unify input_tokens semantics across LLM-producing nodes (#492)

### LINKED ISSUE #492: Inconsistent `input_tokens` semantics (LLMNode vs ClaudeCodeNode) corrupt displayed token totals
## Summary

pflow's two LLM-producing nodes record `input_tokens` with **opposite semantics**, and the two trace consumers each assume a *fixed* interpretation. Displayed **token totals are therefore wrong for one of the two producers** whenever cache tokens are non-zero and/or a workflow mixes `llm` and `claude-code` nodes.

**Scope:** displayed **token counts only**. `cost_usd` is provider-supplied per call (LiteLLM `response_cost` / Claude SDK `total_cost_usd`) and is **not** derived from these token fields, so dollar figures are unaffected.

## Producers disagree on what `input_tokens` means

| Node | `input_tokens` means | Writes `input_token_accounting`? | Writes `uncached_input_tokens`? |
|---|---|---|---|
| `LLMNode` — `src/pflow/nodes/llm/llm.py:1274` | cache-**inclusive** total | yes — `"total_includes_cache"` (`:1286`) | yes (`:1275`) |
| `ClaudeCodeNode` — `src/pflow/nodes/claude/claude_code.py:1443` | **non-cached** input only (explicit `# Only non-cached input tokens`) | **no** | **no** |

## Consumers each pick one interpretation; neither reads the discriminator

| Consumer | Treats `input_tokens` as | Correct for | Wrong for |
|---|---|---|---|
| `src/pflow/core/metrics.py:189` — `"input": sum(input_tokens)`, cache tracked as separate sibling totals | the complete input total | `LLMNode` | `ClaudeCodeNode` (cache omitted from the "input" headline) |
| `src/pflow/core/trace_report.py:350-364` (`_input_token_total` = `input_tokens + cache_creation_input_tokens + cache_read_input_tokens`) | uncached slice only | `ClaudeCodeNode` | `LLMNode` (**double-counts** cache_creation + cache_read) |

Data flows verbatim from producer → `event["llm_call"]` (`runtime/workflow_trace.py`, via `aggregate_llm_usage_with_retries`) → `collect_llm_calls()` → both consumers, with **no harmonization layer**.

## When it manifests

Only when cache tokens > 0 and/or a workflow mixes `llm` and `claude-code` nodes. A single-node-type, no-cache run shows correct numbers — which is why it has gone unnoticed.

## Fix options

The disambiguation contract already exists (`src/pflow/core/llm_usage.py` defines `input_token_accounting`) but is half-wired: `LLMNode` sets it, `ClaudeCodeNode` skips it, and both consumers ignore it.

- **(A)** `ClaudeCodeNode` populates `uncached_input_tokens` + `input_token_accounting="split_cache_fields"` like the LiteLLM adapter, **and** both consumers branch on `input_token_accounting`. Preserves the cache-tier breakdown.
- **(B)** Normalize at the producer boundary so both nodes write `input_tokens` with one agreed meaning, making the discriminator unnecessary. Simpler, but changes a field other code may rely on — **audit all `collect_llm_calls()` readers first** (there may be consumers beyond the two above, e.g. cost / MCP execution paths).

## Provenance

Confirmed by direct read of the cited lines. Discovered during a trace-architecture investigation; `cost_usd` confirmed unaffected (provider-supplied, not token-derived).

spinje:
## Summary
pflow's two LLM-producing nodes recorded `input_tokens` with opposite meanings, so displayed token totals were wrong whenever cache tokens were non-zero or a run mixed `llm` and `claude-code` nodes. This unifies the contract — `input_tokens` is the cache-**inclusive** total at every producer — by fixing the single violating producer (ClaudeCodeNode) and deleting the compensating cache-adding arithmetic in `trace_report.py`.

Fixes #492

## Changes
- **ClaudeCodeNode** emits the inclusive total via a deterministic `_claude_token_fields` helper (used by both `post()` and the schema-retry record), plus `uncached_input_tokens` + `input_token_accounting` for parity with LLMNode's documented `llm_usage` shape.
- **trace_report.py** `_input_token_total` and the summary read `input_tokens` verbatim — no render-time cache re-adding (pure deletion). `workflow_trace.py`'s `total_input_tokens` contract comment flipped to "inclusive".
- **`aggregate_llm_usage_with_retries`** now also sums `uncached_input_tokens`, preserving the `input == uncached + creation + read` invariant on retried claude-code calls.
- **Docs** — `claude-code.mdx` llm_usage example made cache-inclusive and gains `uncached_input_tokens`/`input_token_accounting`; `template-variables.mdx` wording broadened (claude-code is always `split_cache_fields`).
- **Bundled pre-existing fix** (surfaced during review): the CLI/MCP cost line read a non-existent `total_tokens` metrics key (the metrics dict emits `tokens_total`), so the token count silently rendered as 0 — corrected on both surfaces.

## Explanation
Root cause is a provider model mismatch: Anthropic (Claude SDK) reports tokens disjoint-additively (`input_tokens` = uncached only; cache tiers are separate addends), while OpenAI/LiteLLM and pflow's documented contract are inclusive-subset (`input_tokens` = total incl. cache; cache tiers are a subset). LLMNode converts via the LiteLLM adapter; ClaudeCodeNode leaked the disjoint shape. `trace_report` then re-added the cache tiers — correct only for the disjoint shape, double-counting every inclusive LLMNode call. The fix converts at the one producer that knows its provider's semantics (Claude's split is always known, so the inclusive total is a deterministic sum — no value-based heuristic), which leaves all consumers branch-free. `metrics.py` needed no change (it already treats `input_tokens` as the total).

12 files, +476/-64.

## Testing
- `test_claude_token_fields_sums_cache_into_inclusive_total` / `..._defaults_missing_and_none_to_zero` — the helper folds the SDK split into the inclusive total and is None/missing-safe.
- `test_post_emits_inclusive_input_tokens_with_cache` — full prep/exec/post: input 1000 + 25000 cache → `input_tokens=26000`, `total_tokens=28000`, `input_token_accounting=split_cache_fields`; includes an exact key-set shape guard for claude's `llm_usage`.
- `test_schema_retry_aggregates_inclusive_input_tokens_with_cache` — non-zero-cache retry requiring both producer sites inclusive; pins aggregated `input_tokens=1800`, `uncached_input_tokens=300`, and the `uncached + creation + read == input_tokens` invariant (the pre-existing cache-zero retry test could not catch a half-fix).
- `test_metrics_and_report_agree_on_inclusive_input_total` — the bug's real surface: a mixed claude+llm trace run through the real `MetricsCollector` and `WorkflowTraceCollector` asserts metrics `tokens_input` == trace_report rendered "in" (27,000), plus pins `llm_summary.total_tokens`.
- Three trace_report "fold" tests rewritten as inclusive-render tests that double as mutation guards (`cache_creation == input_tokens`, so re-summing would render `3,000` not `1,500`).
- `test_priced_cost_line_shows_token_count` (CLI + MCP, incl. a >1M separator case) — the bundled cost-line fix.
- Mutation-checked: reverting `post()` fails the producer test; reverting only `_usage_record_from` fails the retry test (isolation proven). `make test` 7757 passed; ruff/mypy green in CI.

## Follow-ups
- GH #494 — abbreviate human-facing token counts (`2.1M`/`153K`) via a shared helper (excludes JSON + cache thresholds).
- The claude-code docstring change to `Writes:` subfields only becomes agent-visible after a registry rebuild (`pflow registry scan`); no "unknown parameter" errors (these are outputs, not params).

### DIFF
--- a/docs/how-it-works/template-variables.mdx
+++ b/docs/how-it-works/template-variables.mdx
@@ -258,7 +258,7 @@ Available fields on `${node_id.llm_usage}`:
 | `total_tokens` | int | Input + output |
 | `cache_creation_input_tokens` | int | Tokens used for cache creation |
 | `cache_read_input_tokens` | int | Tokens read from cache |
-| `input_token_accounting` | str | How LiteLLM prompt tokens were normalized (`total_includes_cache` or `split_cache_fields`) |
+| `input_token_accounting` | str | How `input_tokens` was derived. The `llm` node reports `total_includes_cache` or `split_cache_fields` depending on what the provider returned; `claude-code` always reports `split_cache_fields` (the Claude SDK splits cached tokens into separate fields). |
 
 ### Claude Code metadata
 
--- a/docs/reference/nodes/claude-code.mdx
+++ b/docs/reference/nodes/claude-code.mdx
@@ -106,7 +106,7 @@ The node always returns `default` from `post()` — schema soft-failures and SDK
 
 `cost_usd` is the SDK-reported API-equivalent estimated cost. Actual billing depends on auth method; Claude Pro/Max subscription runs may report an API-equivalent cost without a direct per-call charge.
 
-The `retries` field (optional) contains an array of usage records for schema retry attempts. Token/cost values shown at the top level aggregate the main call and all retries — when reading `llm_usage` for cost tracking, use the top-level fields (they already include retry costs).
+The `retries` field (optional) contains an array of usage records for superseded schema retry attempts. The top-level `llm_usage` fields reflect the **final (successful) attempt only** — the superseded attempts live in `retries[]`. The run's reported cost and token totals (the cost summary and the execution trace) aggregate the main call and all retries; the per-node `llm_usage` you read via `${node.llm_usage}` does not.
 
 ## Authentication
 
--- a/docs/reference/nodes/claude-code.mdx
+++ b/docs/reference/nodes/claude-code.mdx
@@ -73,11 +73,13 @@ The node always returns `default` from `post()` — schema soft-failures and SDK
 ```json
 {
   "model": "claude-sonnet-4-5",
-  "input_tokens": 1500,
+  "input_tokens": 21500,
+  "uncached_input_tokens": 1500,
   "output_tokens": 450,
-  "total_tokens": 1950,
+  "total_tokens": 21950,
   "cache_creation_input_tokens": 0,
-  "cache_read_input_tokens": 0,
+  "cache_read_input_tokens": 20000,
+  "input_token_accounting": "split_cache_fields",
   "cost_usd": 0.0234,
   "duration_ms": 3450,
   "num_turns": 2,
@@ -85,9 +87,11 @@ The node always returns `default` from `post()` — schema soft-failures and SDK
   "retries": [
     {
       "input_tokens": 500,
+      "uncached_input_tokens": 400,
       "output_tokens": 150,
       "cache_creation_input_tokens": 0,
       "cache_read_input_tokens": 100,
+      "input_token_accounting": "split_cache_fields",
       "cost_usd": 0.0078,
       "duration_ms": 1200,
       "num_turns": 1,
@@ -97,6 +101,8 @@ The node always returns `default` from `post()` — schema soft-failures and SDK
   ]
 }
 ```
+
+`input_tokens` is the cache-inclusive total — it includes `cache_creation_input_tokens` and `cache_read_input_tokens`. `uncached_input_tokens` is the slice that wasn't served from cache. In the example above, 20,000 of the 21,500 input tokens were read from cache, leaving 1,500 uncached.
 
 `cost_usd` is the SDK-reported API-equivalent estimated cost. Actual billing depends on auth method; Claude Pro/Max subscription runs may report an API-equivalent cost without a direct per-call charge.
 
--- a/src/pflow/cli/workflow_output.py
+++ b/src/pflow/cli/workflow_output.py
@@ -673,15 +673,16 @@ def _format_cost_summary_lines(total_cost: float | None, formatted_result: dict[
     if total_cost is None or total_cost <= 0:
         return []
 
-    # Get token count for context
+    # Get token count for context. The key is ``tokens_total`` (set by
+    # MetricsCollector._build_execution_metrics) — cache-inclusive input + output.
     workflow_metrics = metrics.get("workflow", {})
-    total_tokens = workflow_metrics.get("total_tokens", 0)
+    tokens_total = workflow_metrics.get("tokens_total", 0)
 
     detail_parts: list[str] = []
     if total_llm_calls > 0:
         detail_parts.append(f"{total_llm_calls} call{'s' if total_llm_calls != 1 else ''}")
-    if total_tokens > 0:
-        detail_parts.append(f"{total_tokens:,} tokens")
+    if tokens_total > 0:
+        detail_parts.append(f"{tokens_total:,} tokens")
 
     if detail_parts:
         return [f"💰 Cost: ${total_cost:.4f} ({', '.join(detail_parts)})"]
--- a/src/pflow/core/trace_report.py
+++ b/src/pflow/core/trace_report.py
@@ -348,20 +348,16 @@ def _format_cost(cost: float | None) -> str:
 
 
 def _input_token_total(llm_call: dict[str, Any]) -> tuple[int, int]:
-    """Return ``(total_input_tokens, cache_read_tokens)`` for one LLM call.
+    """Return ``(input_tokens, cache_read_tokens)`` for one LLM call.
 
-    LLM input is billed in three tiers that SUM to the true input the model
-    processed: uncached (``input_tokens``) + cache-write
-    (``cache_creation_input_tokens``) + cache-read (``cache_read_input_tokens``).
-    The report headlines that total and surfaces the cache-read share as a
-    cache-hit %, rather than showing the uncached slice alone \u2014 which made
-    cache-heavy agent runs look like ``in`` << ``out`` and divorced the cache
-    numbers from the token line.
+    ``input_tokens`` is pflow's cache-INCLUSIVE input total (see
+    ``core/llm_usage.py``) \u2014 every producer (LLMNode and ClaudeCodeNode) emits it
+    that way, so the report headlines it directly with no render-time cache
+    arithmetic. ``cache_read`` is returned only for the cache-hit % display.
     """
-    uncached = llm_call.get("input_tokens", llm_call.get("prompt_tokens", 0)) or 0
-    cache_write = llm_call.get("cache_creation_input_tokens", 0) or 0
+    total_in = llm_call.get("input_tokens", llm_call.get("prompt_tokens", 0)) or 0
     cache_read = llm_call.get("cache_read_input_tokens", 0) or 0
-    return uncached + cache_write + cache_read, cache_read
+    return total_in, cache_read
 
 
 def _format_tokens(total_in: int, tokens_out: int, cache_read: int, *, with_cache_pct: bool) -> str:
@@ -798,10 +794,11 @@ def _build_summary(
                 llm.get("total_num_turns", 0),
             )
         )
-        # `in` is the TRUE total: uncached + cache-write + cache-read (the three
-        # billing tiers). total_input_tokens alone is just the uncached slice.
+        # total_input_tokens is the cache-INCLUSIVE input total (every producer
+        # emits inclusive input_tokens; see core/llm_usage.py). cache_read is read
+        # only for the cache-% suffix, NOT added back into the input total.
         cache_read = llm.get("total_cache_read_tokens", 0)
-        total_in = llm.get("total_input_tokens", 0) + llm.get("total_cache_creation_tokens", 0) + cache_read
+        total_in = llm.get("total_input_tokens", 0)
         tokens_out = llm.get("total_output_tokens", 0)
         if total_in or tokens_out:
             lines.append(f"- Tokens: {_format_tokens(total_in, tokens_out, cache_read, with_cache_pct=True)}")
--- a/src/pflow/execution/formatters/success_formatter.py
+++ b/src/pflow/execution/formatters/success_formatter.py
@@ -356,14 +356,16 @@ def format_success_as_text(  # noqa: C901
         if total_llm_calls > 0:
             lines.append(f"   Total LLM calls: {total_llm_calls}")
     elif total_cost and total_cost > 0:
+        # The key is ``tokens_total`` (set by MetricsCollector._build_execution_metrics)
+        # — cache-inclusive input + output. Mirrors the CLI cost line.
         workflow_metrics = metrics.get("workflow", {})
-        total_tokens = workflow_metrics.get("total_tokens", 0)
+        tokens_total = workflow_metrics.get("tokens_total", 0)
 
         detail_parts: list[str] = []
         if total_llm_calls > 0:
             detail_parts.append(f"{total_llm_calls} call{'s' if total_llm_calls != 1 else ''}")
-        if total_tokens > 0:
-            detail_parts.append(f"{total_tokens:,} tokens")
+        if tokens_total > 0:
+            detail_parts.append(f"{tokens_total:,} tokens")
 
         if detail_parts:
             lines.append(f"💰 Cost: ${total_cost:.4f} ({', '.join(detail_parts)})")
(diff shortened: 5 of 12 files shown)


spinje:
# Progress Log — Fix #492: Unify `input_tokens` semantics

Plan: `~/.claude/plans/yes-lets-write-the-zesty-corbato.md`
Branch: `fix/input-tokens-semantics`

## Goal (one line)
Make `input_tokens` the cache-**inclusive** total at every LLM producer (fix
ClaudeCodeNode, the lone violator) and delete the redundant render-time cache-adding
in `trace_report.py`, so `metrics.py` and `trace_report.py` agree for both node types.

## Verification of plan against code (done before coding)
- `metrics.py:189` sums `input_tokens` as total — confirmed, no change. ✅
- `claude_code.py:1435-1455` (`post`) leaks Anthropic disjoint `input_tokens`. ✅
- `claude_code.py:787-797` (`_usage_record_from`) likewise. ✅
- `trace_report.py:350-364` (`_input_token_total`) adds cache tiers → double counts. ✅
- `trace_report.py:801-804` summary = three-term sum. ✅
- `workflow_trace.py:316-318` contract comment says "UNCACHED slice". ✅
- `workflow_trace.py:597,621,627` aggregator sums per-attempt `input_tokens`, recomputes total. ✅
- `_make_event` default node_type = "TestNode" (fold tests key off `llm_call` presence, fine). ✅
- fixture builder `cached_llm_event_with_call` default `input_tokens=1000, cache_read=950`. ✅
- `claude_code.py` imports: `from typing import Any, Optional` — need `Mapping`. ✅

## Implementation steps
- [x] 1. claude_code.py: add `_claude_token_fields` helper + `Mapping` import
- [x] 1b. rewire `post()`
- [x] 1c. rewire `_usage_record_from()`
- [x] 1d. interface docstrings (2 copies: ~14-18, ~159-163)
- [x] 2. trace_report.py: `_input_token_total` + summary deletion
- [x] 3. workflow_trace.py: flip contract comment (no code)
- [x] 4. docs (claude-code.mdx, template-variables.mdx)
- [x] 5. tests (producer, consumer, cross-consumer, rewrites, fixtures)
- [x] verify: targeted pytest, make test, make check, mutation checks

## Deviation: total_tokens pin location
Plan §5/§Verif-4 said "add a total_tokens assertion to the summary fixture
(test_agent_calls_label_with_turns_and_cache)". But `_build_summary` never renders
`total_tokens` (grep-confirmed: zero occurrences of total_tokens in trace_report.py),
so asserting it on that test's rendered output is impossible, and asserting it on the
hand-built fixture dict would be tautological + a test-fidelity violation (#19).
Instead pinned `llm_summary.total_tokens` (= 29500) in the cross-consumer test, where it
is computed by the REAL `_collect_llm_summary` aggregation over inclusive per-call
`total_tokens` — the location with actual teeth against drift. Plan intent honored,
better placement.

## Learnings / deviations

### Outcome
All steps implemented. `make test` 7753 passed / 1 skipped; `make check` clean
(ruff, ruff-format, mypy 230 files, deptry). Both mutation checks confirmed teeth:
- revert `post()` input_tokens → `test_post_emits_inclusive_input_tokens_with_cache` fails (200 ≠ 26000).
- revert only `_usage_record_from` → `test_schema_retry_aggregates_inclusive_input_tokens_with_cache`
  fails (retries[0] = 100 ≠ 600) while the producer test stays green. Isolation proven.

### Key insights
1. **The fix is a deletion, not a branch.** Both producers now emit inclusive
   `input_tokens`; the consumer (`trace_report`) stops re-adding cache. No
   `input_token_accounting` branching anywhere — that field records *provenance*, not
   *how to read the normalized output*. Confirmed `metrics.py` needed zero change.
2. **The summary-path bug and per-call-path bug are independent code.** `_build_summary`
   reads `total_input_tokens` directly (does NOT call `_input_token_total`). Both had the
   three-term sum; both fixed. Tests cover both paths separately.
3. **`_claude_token_fields` is deterministic, unlike the LiteLLM normalizer.** Claude's
   split is always known so `total = uncached + creation + read` is exact — no value-based
   `prompt >= cacheable` heuristic (which would misfire for claude when uncached ≥ cacheable).

### Deviations (with rationale)
- **`total_tokens` pin moved off the summary fixture** (detailed above): `_build_summary`
  never renders `total_tokens` (grep: 0 hits in trace_report.py), so the plan's literal
  placement was un-assertable and a hand-built-fixture assertion would be tautological
  (tests/CLAUDE.md #19). Pinned it in the cross-consumer test against the REAL
  `_collect_llm_summary` aggregation instead — same intent, real teeth.
- **`post()` uses `**token_fields` + `token_fields["input_tokens"]` for `total_tokens`**
  rather than re-reading a local. The plan's mutation-check wording assumed a single
  `input_tokens =` line per site; the spread is cleaner (one source of truth) and the
  mutation checks still isolate each site (verified above).
- **Fold-test rewrites keep `cache_creation == input_tokens`** (not just adding an
  `input_tokens` key). This turns each rewrite into a live mutation guard: if a future
  edit re-sums cache, `in` renders 3,000 not 1,500 — the test fails loudly. Pure
  semantic rewrite per tests/CLAUDE.md #18.
- **`uncached_input_tokens` / `input_token_accounting` already existed in the shared
  template-variables.mdx table** (inclusive description). Only line 261's LiteLLM-specific
  wording needed broadening — claude-code was the doc-spec violator the plan flagged, now
  aligned. Added the two fields to both claude_code.py docstring copies + claude-code.mdx JSON.

### Plan factual error found during verification (display-path pin NOT added — with reason)
Plan §Files-touched + §Verification-4 claim `cli/workflow_output.py:678` and
`success_formatter.py:360` "re-derive tokens_input + tokens_output ... which auto-corrects"
and asked to pin the live summary "N tokens" line. **This premise is wrong.** Both sites
read `workflow_metrics.get("total_tokens", 0)`, but `_build_execution_metrics`
(metrics.py:236) emits the key `tokens_total`, NOT `total_tokens`. Empirically confirmed:
the workflow metrics dict has no `total_tokens` key, so both reads return 0 and the
"💰 Cost: $X (N calls, M tokens)" line NEVER shows the tokens detail. No test asserts it.
→ This is a PRE-EXISTING key-name bug, orthogonal to #492 (token *semantics*). I did NOT
add the §4 "N tokens" pin because there is nothing truthful to pin (always 0), and the plan
explicitly says "no code change" for these files. The token *values* that ARE displayed
(trace_report summary "Tokens: N in / M out", per-node report, JSON metrics
tokens_input/output/total) are all inclusive and pinned by the cross-consumer test.
Surfaced to user as an out-of-scope decision (changes user-visible output → needs confirm).

### Follow-up fix applied (user-approved): cost-line token-count key mismatch
Fixed the pre-existing bug found above. `cli/workflow_output.py` + `success_formatter.py`
now read `workflow_metrics.get("tokens_total", 0)` (was `total_tokens` → always 0). Renamed
the local to `tokens_total` to mirror the source key (the bug was a name divergence; aligning
prevents recurrence). Both display fixtures (`test_direct_execution_helpers.py`,
`test_success_formatter.py`) corrected to the production key `tokens_total` (they encoded the
bug-compatible `total_tokens` — the tests/CLAUDE.md #19 trap; no assertion read it so the dead
detail was invisible). Added CLI + MCP parity tests asserting the token count renders, incl. a
>1M case (`2,137,122 tokens`).

**>1M behavior (user asked):** `f"{tokens_total:,} tokens"` → comma-grouped, no overflow
(Python int), no abbreviation. Deliberately matches the trace-report summary convention
(`Tokens: N in / M out` also uses `:,`); abbreviating only here would create cross-surface
inconsistency. Mutation-checked: reverting the key → `💰 Cost: $0.0500 (9 calls)` (tokens
gone) → both new tests fail. make test 7757 passed; make check clean.

### Follow-up issue filed: token abbreviation
GH #494 (spinje/pflow) — abbreviate human-facing token counts (`2.1M`/`153K`) via one
shared `format_token_count` helper. Scoped to exclude JSON (agents need exact ints) and
cache thresholds/minimums (precise). Decided AGAINST abbreviating the cost line alone here
(cross-surface inconsistency); deferred to #494 to do it everywhere at once.
Body: `scratchpads/fix-492-input-tokens-semantics/issue-body.md`.

### Notes for downstream
- Docstring change to claude-code's `Writes:` subfields only becomes agent-visible after a
  registry rebuild (`pflow registry scan`) — the context builder serves the cached
  `~/.pflow/registry.json`. No "unknown parameter" errors (these are outputs, not params).
- S105 false-positive on `assert lu["input_token_accounting"] == "split_cache_fields"`
  (ruff reads the var name's "token" as secret-ish) — scoped `# noqa: S105` with reason.

## Review follow-up (post-implementation review)

### Finding 1 (FIXED): retry aggregator left `uncached_input_tokens` stale
`aggregate_llm_usage_with_retries` summed `input_tokens` + cache tiers across attempts but
NOT `uncached_input_tokens` (it carried the main attempt's value via the shallow copy). The
fix newly makes claude emit `uncached_input_tokens`, so on a *retried* claude call the
aggregated trace `llm_call` had `input_tokens` summed (1800) but `uncached` main-only (200)
— breaking the invariant `input == uncached + creation + read` on that dict. Harmless today
(no consumer reads `uncached` off the aggregated trace dict; templates read the consistent
shared `post()` dict), but a latent trap the fix introduced.
- **Fix**: sum `uncached_input_tokens` in the aggregator's main-init + retry loop
  (`workflow_trace.py`). Pinned in `test_schema_retry_aggregates_inclusive_input_tokens_with_cache`
  (`agg["uncached_input_tokens"] == 300` + explicit invariant assertion). Existing
  `test_aggregate_llm_usage.py` unaffected (field-by-field asserts; no-retry path untouched).
  315 targeted + 144 smoke tests pass; ruff clean.

### Finding 2 (RETRACTED — review-sandbox artifact, NOT a real issue)
Initial review flagged 8 mypy errors at `claude_code.py:80-83` (`CLINotFoundError = None`,
the pre-existing SDK exception import fallback). **This was a false alarm.** Verified:
- The fallback is pre-existing (stash-confirmed) and untouched by #492.
- **CI is GREEN on the base commit `4d76ff64`**: the `quality` job log shows mypy 1.19.1 →
  "Success: no issues found in 230 source files", and mypy gates CI (no continue-on-error).
  The implementer's `make check` was green too.
- The errors reproduce ONLY in the reviewer's macOS sandbox; CI runs Linux with an otherwise
  identical toolchain (mypy 1.19.1, claude-agent-sdk 0.2.82, Python 3.13, `uv sync --frozen`).
  The platform difference changes how mypy resolves claude-agent-sdk's `._errors` exception
  types (concrete `type[X]` on macOS → `= None` flagged; not flagged on the CI install).
- **Nothing to fix.** Not a code defect, not introduced by #492, not seen by CI or the user.
  The earlier "warn_unused_ignores / SDK-version-dependent" reasoning was speculation the CI
  evidence supersedes.

## PR review round 2 (automated bots on PR #495)

Two automated reviews landed on the PR; evaluated both via `/evaluate-review` (read the
code + parallel `pflow-codebase-searcher` verification of every load-bearing claim).
Outcome: **3 confirmed (all low severity), 1 disputed, 2 acknowledged-no-action.** Confirmed
items fixed in follow-up commit `a42506ba` (`[skip review]`).

### [G1] FIXED — None-guard `output_tokens` in `post()` (Gemini, high)
`total_output = usage.get("output_tokens", 0)` returns `None` when the SDK reports
`output_tokens: None` (the `, 0` default only fires on ABSENT keys), then
`token_fields["input_tokens"] + total_output` → `TypeError`. Real gap, and the lone
exception to the codebase's `or 0` convention (the helper + `aggregate_llm_usage_with_retries`
both guard). Fix: `... or 0`. Regression test `test_post_coerces_none_output_tokens_to_zero`
(mutation-checked: without `or 0`, `input + None` raises).

### [G2] FIXED — same guard in `_usage_record_from()` (Gemini, medium)
`"output_tokens": usage.get("output_tokens", 0)` → `... or 0`. Milder (the aggregator already
guards downstream so no crash), but keeps the retry record type-clean + consistent with G1.

### [S1] FIXED — corrected the `retries` doc prose (strong review, suggestion)
`claude-code.mdx` claimed "top-level fields aggregate the main call and all retries … already
include retry costs." Verified FALSE: `post()` writes top-level from the FINAL attempt only
(`cost_usd = metadata.get("total_cost_usd")`; superseded attempts in `retries[]`); aggregation
happens downstream in `aggregate_llm_usage_with_retries` (the trace/metrics path, NOT
`${node.llm_usage}`). The PR's own example (top-level 21,500 + a 500 retry) contradicted the
prose. Reworded: top-level = final attempt; the run's cost/token summary + trace aggregate all
attempts.

### [S2] DISPUTED — `local_cache_input_tokens` does NOT shift
Reviewer (informational): inclusive `input_tokens` grows this analyzer sum. Verified moot —
claude-code events never carry `cache_source` (gated to LLMNode only via
`_should_write_cache_metadata`, independent of `cache: true`), and the sum only counts events
with `cache_source in {memo, in_process}` (`trace_loading.py:874`). The only feeders are
LLMNode hits, whose `input_tokens` was already inclusive pre-PR. No behavior change.

### Acknowledged — no action
- **`has_cache_telemetry` asymmetry** (claude-code omits it; LLMNode emits it): pre-existing,
  intentional, no claude-code consumer. Documented as deliberate in the plan.
- **Old traces under-report `in`**: accepted under the no-backward-compat policy.

### Verification (round 2)
114 claude tests pass (incl. the new None-guard test); ruff clean; all pre-commit hooks green
(incl. the MDX check). Follow-up commit `a42506ba` pushed with `[skip review]`.

Issue #484, opened 2026-06-06
Batch node `errors` output is `None` when empty — should always be a list (`[]`)


spinje:
## Summary

A batch node's `errors` output resolves to **`None` when there are no failures**, instead of an empty list `[]`. It is the only batch output field that collapses to `None` — its siblings (`results`, `error_count`) use their natural empty values (`[]`, `0`). This breaks the field's own declared type (`array`), is inconsistent within the same output dict, and is a footgun for downstream consumers and AI agents authoring workflows.

**Fix:** make batch `errors` always a list (`[]` when empty), aligning runtime with the already-declared `type: "array"` contract.

## Scope guard — read first

This issue is **only** about the batch-node aggregate output field at `shared[node_id]["errors"]` (the per-item failure record produced by `_aggregate_batch_results`). 

It is **NOT** about any of these other, unrelated `errors` fields (do not touch them):
- `ExecutionResult.errors` / `ValidationResult.errors` (properties in `execution/result.py`)
- `formatted["errors"]` (output of `execution/formatters/error_formatter.py`)
- `vresult.errors` (validation results)
- `metadata.get("errors")` / `sdk_errors` (Claude SDK, `nodes/claude/claude_code.py`)
- `__template_errors__`, `template_errors` (engine template resolution)

A repo-wide grep for `errors` returns mostly these unrelated fields. Stay scoped to the batch aggregate output.

## What / how it manifests

A downstream `code` node that consumes a batch node's `errors` output and annotates it with the obvious/declared type fails at runtime:

```markdown
### parse
- type: code
- inputs:
    errors: ${fetch.errors}     # fetch is a batch node with error_handling: continue
```
```python code
errors: list                    # the documented type — type: "array"
...
```

When the batch had **zero failures**, `${fetch.errors}` resolves to `None`, and the code node's input type validation rejects it:

```
Input 'errors' expects list but received NoneType
```

The only current workaround is to annotate `errors: Any` and guard with `errors or []` everywhere downstream — a guard required purely because the producer returns `None` instead of `[]`.

## Root cause

`src/pflow/runtime/engine/batch_executor.py`, in `_aggregate_batch_results` (~line 1004), the output dict written to `shared[node_id]`:

```python
shared[node_id] = {
    "results": successful_results,          # always a list ([] when empty)
    "count": len(exec_res),                 # always int
    "success_count": success_count,         # always int
    "error_count": len(errors),             # always int (0 when none)
[...]

PR #505, opened 2026-06-11
refactor: lift planner's engine-mirror into shared runtime functions (#504)

### LINKED ISSUE #504: Dry-run planner hand-mirrors engine walk semantics — lift the mirror into shared functions
## Problem

`execution/plan.py` (the dry-run planner, ~2,500 lines) re-implements runtime-engine semantics by hand. The shared `plan_node()` seam (Task 156) covers cache/template/hash decisions, but everything *around* it is a hand-maintained mirror held in sync by comments — and the record shows the mirror leaks:

- **8–9 of 10 commits** since the dry-run feature landed touched `plan.py` in lockstep with engine changes; the one exception was itself a 673-line parity-bug fix.
- **4 parity bugs shipped to main**: #318/#323 (batch sub-workflow recursion missing), #333 (cache-key serialization drift), #336 (heterogeneous batch dry-run).
- The **Task 162 cost under-report** passed the entire 2,290-line drift suite undetected and was caught only by manual review — the drift suite pins control-flow parity, not the mirrored surfaces.
- The standing invariant from the Task 162 review: *"Adding any new engine walk-state key means mirroring it in `create_planner_shared`, or the dry-run diverges silently."*

Tasks 164 (resume from failed node) and 125 (HITL gates) are next on the roadmap and both touch this mirror surface — each will pay the lockstep tax unless the mirror moves behind the seam first.

## Mirrored surfaces

1. **`__execution__` dict literal** — byte-identical in 3 places: `instrumentation.py::initialize_execution_state`, `plan.py::create_planner_shared`, `node_state.py::mark_node_failed` (defensive init).
2. **`_validate_only_target`** — near-verbatim duplicate in `engine.py` and `plan.py`. The copies have already drifted on one edge: `--only ""` silently falls through to a **full run** on the engine side but raises `CompilationError` on the planner side.
3. **Batch output shape** — the 6-key output dict (+ 6-key `batch_metadata`) is a copied literal in `batch_executor.py::_aggregate_batch_results` and `plan.py::_build_batch_output_shape` (the #484 `errors: None→[]` fix had to be applied to both by hand).
4. **Routing precedence** (successor match > clean termination > routing error) — exists once in the engine walk loop + `_handle_no_successor`, once in `plan.py::_classify`. `is_clean_termination` is already shared; the precedence rule is not.
5. **Degraded-snapshot advisory** — the `only.snapshot-degraded` Diagnostic is constructed twice (engine + planner), identical except verb tense.

## Proposal

Extend the proven house pattern (`plan_node`, `is_clean_termination`, `loop_runtime_scope`, …): extract each surface into a shared pure function in `runtime/`, with the engine and the planner as its two callers, docstrings naming both consumers.

**Explicitly NOT in scope** (per ADR-0001 and the ADR-0006 doctrine — share data structures and rules, not traversal): merging the walkers, any WalkEnvironment/traversal abstraction, the prompt-cache install ritual, `__loop_active__` propagation, routing-error message tenses.

## Acceptance criteria

- Pure refactor: `test_plan_drift.py`, `test_loop_plan.py`, `test_plan_classify.py`, `test_engine_behavior.py`, `test_failed_node_invariant.py` pass **unmodified**.
- The `--only ""` divergence is resolved explicitly (recommend the planner's loud error — a silent full run on malformed agent input is the failure class this project forbids) and pinned by a test.
- Focused unit tests for each shared function.
- A call-site parity test for the batch output shape (engine's actual `shared[node_id]` vs the planner's prediction) — the drift suite demonstrably does not catch a re-forked shape literal.

spinje:
## Summary

Extracts five hand-mirrored engine/planner surfaces into shared pure functions in `runtime/`, so the dry-run planner and the engine become two callers of one implementation — parity by construction instead of by "keep in lockstep" comments. This lands ahead of Tasks 164 (resume) and 125 (HITL gates), both of which touch this mirror surface and would otherwise pay the lockstep tax.

Fixes #504

## Changes

- **`new_execution_state()`** (`node_state.py`) — the canonical 5-key `__execution__` dict, previously byte-identical in 3 places (`instrumentation.py`, `plan.py`, `node_state.py`)
- **`validate_only_target()`** (`engine.py`) — one parse-and-validate for `--only`, previously near-verbatim in engine + planner
- **`build_batch_output()`** (`batch_executor.py`) — single source of the 6-key batch output contract + 6-key `batch_metadata`; the planner passes the no-execution degenerates (`errors=[]`, `timing_stats=None`)
- **`route_action()`** (`engine.py`) — the routing precedence kernel (successor match > clean termination > routing error) expressed once; the engine maps `ROUTING_ERROR` to its side effects, the planner overlays its BOUNDARY concept
- **`build_snapshot_degraded_diagnostic()`** (`engine.py`) — one construction site for the `only.snapshot-degraded` advisory (verb tense / source / context are the per-surface parameters)
- Doc updates: `engine/CLAUDE.md`, `execution/CLAUDE.md`, ADR-0002 stale pointer, plus the stale `workflow_executor.py` line-number comment in `plan.py`

## Explanation

The drift record motivating this (full evidence in #504): 8–9 of 10 post-dry-run commits touched `plan.py` in lockstep with engine changes, 4 parity bugs shipped to main, and the Task 162 cost under-report passed the entire 2,290-line drift suite — every drift bug lived in the mirror surface *outside* the shared `plan_node()` seam. This PR applies the existing house pattern (shared pure functions with docstrings naming both consumers) to the remaining mirror sites. Deliberately NOT a walker merge — per ADR-0001 and the ADR-0006 doctrine, it shares data structures and rules, never the traversal.

**One intentional behavior change**: `--only ""` previously diverged between the copies — the engine silently fell through to a **full run**, the planner raised. The unified function adopts the loud error (`Node '' not found` + available nodes) on both surfaces; a silent full run on malformed agent input is exactly the failure class this project forbids.

One honest deviation surfaced by mypy: the engine's `exec_res` can carry `None` placeholders (parallel pre-sizing / fail-fast) that error-index filtering doesn't remove — the shared builder's parameter is typed `Sequence[Optional[dict]]` and documents the pass-through, preserving the pre-extraction literal's behavior exactly rather than silently filtering.

Net: 14 files, +660/−190 (the +660 is dominated by new tests and consumer-naming docstrings; `plan.py` loses ~100 lines of mirror code and ~6 lockstep comments die).

## Testing

- **Parity suites pass unmodified** — `test_plan_drift.py`, `test_loop_plan.py`, `test_plan_classify.py`, `test_engine_behavior.py`, `test_failed_node_invariant.py` were the designated revert-triggers; zero edits needed.
- `test_plan_batch_sub_workflow_output_shape_matches_engine` (new, drift suite) — runs a batch sub-workflow for real AND plans it, asserting the engine's actual `shared[node_id]` structurally equals the planner's prediction (key order both levels, exact values, per-item stamping). **Mutation-verified**: a re-forked planner literal with drifted keys (`errors: None` — the #484 regression — plus 3 dropped `batch_metadata` keys) passes all 43 pre-existing batch/drift tests and fails only this one.
- `TestValidateOnlyTarget::test_empty_string_raises` — pins the one intentional behavior change
- `test_mark_node_failed_defensive_init_has_no_cache_hits` — pins the deliberate asymmetry (defensive init must NOT gain `__cache_hits__`)
- `test_calls_share_no_mutable_objects` — guards against a future module-level-constant/shallow-copy "optimization" of the factory
- New `test_route_action.py` (8 tests) and `test_batch_output_shape.py` (9 tests) pin the kernels' contracts directly
- `test_snapshot_degraded_diagnostic_builder_shares_identity_across_sources` — both surfaces share id/title/suggestions; messages differ exactly by the verb
- Full: `make test` (7,796 passed), `make test-e2e` (43 passed), `make check` clean
- Manual CLI verification: basic walk, batch + downstream `${node.batch_metadata.*}` templates, `--dry-run`, routing-error path (exit code + archived failure + hint), the full `--only` matrix (valid/empty/typo/dotted, run AND plan), and a condition-terminated loop (3 re-entries through the rewritten walk loop)

## Follow-ups (filed separately, deliberately out of this PR's scope)

- #506 — the drift suite's **cost-parity blind spot**: this PR pins control-flow and shape parity, but plan-predicted cost vs engine-actual cost remains unpinned (how the Task 162 under-report hid). None of this PR's extractions touch cost math.
- Noticed and preserved verbatim (documented in `build_batch_output`'s docstring, not filed): `None` placeholders in batch `results` under parallel fail-fast pass through as the pre-extraction literal did — fixing it here would have violated the zero-behavior-change contract.

---
Implemented by Claude (Session: e127fea8-6a39-4845-8e99-329817fd5d04)


spinje:
Review findings addressed in 29049e7f:

- **[W1] `child_only` always `None`** — documented in `validate_only_target`'s docstring: dotted targets raise above; the tuple shape is reserved for the deferred #443 nested-targeting follow-up.
- **[S2] Import hoist** — taken, and it paid extra: the module-level `Diagnostic`/`Severity` import retired not just the builder's lazy import + `TYPE_CHECKING` ceremony but two additional pre-existing function-local imports of the same names.
- **[S4] `Optional[str]` → `str | None`** — matched to the adjacent `parse_only_path`.
- **[S5] `runtime/CLAUDE.md` reserved-keys block** — annotated with `node_state.new_execution_state()` as the canonical constructor (kept the full reference block since it documents the engine-only keys too).

**Deferred — [S3] `_classify` consuming `RouteDecision.next_node`**: agreed with your own assessment that the smaller diff was the right trade here. Threading `next_node` through `Decision` widens the walker's state for a duplicate-lookup micro-optimization in a once-per-dry-run path. Earmarked for whenever `_advance` is next touched (likely Task 125's gate transition).

`make check` clean, full suite 7,796 passed.

PR #515, opened 2026-06-15
fix: core-node refresh no longer wipes synced MCP tools (#462)

### LINKED ISSUE #462: Core-node auto-refresh wipes all synced MCP tools (preservation filter misses MCP nodes — they have no `type`)
## Summary

Every synced MCP tool is silently dropped from the registry whenever pflow's core-node auto-refresh fires. The refresh is *meant* to preserve MCP nodes (`_refresh_core_nodes` keeps `type in ("user", "mcp")`), but **MCP node entries never get a `type` field**, so the preservation filter skips them and they're wiped. User nodes survive (they're stamped `type="user"`); MCP nodes don't.

The auto-refresh triggers on any change to node source-file mtimes, so anyone **developing pflow** (editing node files, switching branches/worktrees, or letting `uv` reinstall the package) loses all their MCP tools on the next command and has to re-run `pflow mcp sync --all`. End users who never touch the source rarely hit it.

## Environment

- pflow `0.13.0` (commit `e9bf032a`)
- macOS, running via `uv run --project <pflow checkout> pflow`

## Reproduction

```bash
pflow mcp sync --all
pflow mcp list            # e.g. resend (77 tools), playwright (29 tools)  ✅

# Simulate a dev edit / what `uv` reinstall does — bump a node source mtime:
touch src/pflow/nodes/python/python_code.py

pflow mcp list            # every server now: "0 tools ... Run: pflow mcp sync"  ❌
```

No content change is needed — touching the mtime of a single file under `src/pflow/nodes/` is enough. The MCP tools are gone until the next `pflow mcp sync --all`.

## Root cause

**1. The refresh trigger.** On load, the registry refreshes core nodes when the stored version differs from the running pflow version **or** when any node source file is newer than `last_core_scan` (`src/pflow/registry/registry.py:412`, `_source_newer_than_scan`, which walks `pflow.nodes/**.py` mtimes at lines ~446–451). Active development changes those mtimes constantly, so the refresh fires on nearly every command.

**2. The preservation bug.** `_refresh_core_nodes` re-discovers core nodes and merges back the non-core ones (`src/pflow/registry/registry.py:464`):

```python
preserved = {name: data for name, data in nodes.items() if data.get("type") in ("user", "mcp")}
```

But MCP node entries in `registry.json` have **no `type` field**. Compare a core node vs an MCP node:

```
core node 'llm':
  keys: [class_name, docstring, file_path, interface, module, type]
  type = 'core'
  file_path = <.../src/pflow/nodes/llm/llm.py>

MCP node 'mcp-resend-send-email':
  keys: [class_name, file_path, interface, module]     # <-- no 'type'
  type = None
  file_path = 'virtual://mcp'
  module = 'pflow.nodes.mcp.node'
```

So `data.get("type")` is `None` for every MCP node → not in `("user", "mcp")` → not preserved → dropped. The registrar builds the entry with the `virtual://mcp` sentinel but never stamps a type (`src/pflow/mcp/registrar.py:261`, also `src/pflow/nodes/mcp/node.py:54`), whereas user nodes *are* stamped (`src/pflow/registry/registry.py:602` → `node["type"] = "user"`). MCP is the only category the predicate names but never actually sets.

## Suggested fix

**Root cause** — stamp the type when registering MCP nodes, mirroring user nodes. In `src/pflow/mcp/registrar.py` (~line 261), where the node dict is built with `"file_path": "virtual://mcp"`, add:

```python
"type": "mcp",
```

(A single `pflow mcp sync --all` after this re-stamps existing entries.)

**Hardening** — make the preservation robust to registries synced before the fix, in `src/pflow/registry/registry.py:467`:

```python
preserved = {
    name: data
    for name, data in nodes.items()
    if data.get("type") in ("user", "mcp") or data.get("file_path") == "virtual://mcp"
}
```

Either alone fixes the wipe; together they also self-heal existing registries. A regression test that registers an MCP node, triggers a core refresh (bump `last_core_scan` into the past or touch a node file), and asserts the MCP node survives would lock it in.

## Workaround

`pflow mcp sync --all` after any pflow rebuild / branch switch / source edit.

spinje:
## Summary

Every synced MCP tool was silently dropped from the registry whenever pflow's core-node auto-refresh fired. The refresh *intended* to preserve MCP nodes (it kept `type in ("user", "mcp")`), but **MCP entries were never stamped with a `type`**, so the preservation filter skipped them and wiped them. User nodes survived (they're stamped `type="user"`); MCP nodes didn't.

The refresh triggers on any change to node source-file mtimes, so anyone **developing pflow** (editing node files, switching branches/worktrees, `uv` reinstalls) lost all MCP tools on the next command and had to re-run `pflow mcp sync --all`.

Fixes #462

## Changes

- **`src/pflow/mcp/registrar.py`** — stamp `"type": "mcp"` on MCP registry entries in `_create_registry_entry`, mirroring how core (`type="core"`) and user (`type="user"`) nodes are stamped.
- **`src/pflow/registry/registry.py`** — change `_refresh_core_nodes` preservation from an allowlist to the fail-safe denylist `data.get("type") != "core"`, plus a docstring/comment explaining the rationale.
- **`tests/test_mcp/test_mcp_discovery_critical.py`** — assert the registrar now stamps `type="mcp"`.
- **`tests/test_registry/test_registry.py`** — add an untyped legacy `virtual://mcp` entry to the refresh-preservation test and assert it self-heals.

```
 src/pflow/mcp/registrar.py                    |  1 +
 src/pflow/registry/registry.py                | 14 +++++++++++---
 tests/test_mcp/test_mcp_discovery_critical.py |  2 ++
 tests/test_registry/test_registry.py          | 19 +++++++++++++++----
 4 files changed, 29 insertions(+), 7 deletions(-)
```

## Explanation

There were two ways to harden the preservation, and the denylist was chosen deliberately over the more obvious "add a `virtual://mcp` sentinel clause to the allowlist":

The original allowlist (`type in ("user", "mcp")`) failed **unsafe** — it silently dropped anything it didn't recognize, which is exactly what caused this bug. `_refresh_core_nodes` re-discovers *core* nodes, so the true intent is simply "keep everything that isn't core." Expressed as `type != "core"`, the predicate:

- fails **safe** — an unrecognized entry is preserved, never silently dropped (fixes the failure *mode*, not just this instance);
- **self-heals** legacy registries — MCP entries synced before this fix have no `type`, and `None != "core"`, so they survive with no re-sync needed;
- is the simplest final code — one condition, no magic string, no migration crutch to remove later.

Stamping `type="mcp"` is then complementary: it isn't required for the fix (the denylist preserves typed and untyped entries alike), but it makes the registry data self-describing — every node category carries an explicit `type`, so the next reader doesn't have to learn that "MCP is the typeless category identified by a `virtual://mcp` path." This also makes the existing `registry/CLAUDE.md` "Node Types" table (which already documented MCP getting `type` from the registrar) accurate.

## Testing

- `test_create_registry_entry_structure` — extended to assert `entry["type"] == "mcp"`, locking the registrar stamp.
- `test_refresh_preserves_user_nodes` — extended with an untyped legacy `virtual://mcp` entry (the exact shape that was being wiped); asserts it survives a real `_refresh_core_nodes`, locking the denylist self-heal.
- Full suite: **7820 passed, 0 failed**; `make check` clean (ruff, mypy on 230 files, deptry).
- **Manual, real CLI, real mtime trigger path** (isolated `$HOME`, version matched so `_source_newer_than_scan` is what fires):
  - Old code, 2 untyped MCP entries → `pflow mcp list` shows `0 total across 0 servers` (bug reproduced).
  - New code, identical seed → `pflow mcp list` shows `2 total across 1 servers` (fixed). Disk inspection confirmed the refresh genuinely ran (`last_core_scan` moved off `2020`, 11 core nodes re-discovered) and both untyped entries persisted.
  - A normal workflow (`pflow <wf>.pflow.md`) still compiles, executes, and outputs correctly against the refreshed registry.

### DIFF
--- a/src/pflow/mcp/registrar.py
+++ b/src/pflow/mcp/registrar.py
@@ -259,6 +259,7 @@ class MCPRegistrar:
             "class_name": "MCPNode",
             "module": "pflow.nodes.mcp.node",
             "file_path": "virtual://mcp",  # Virtual path for MCP nodes
+            "type": "mcp",  # Mirrors core/user nodes — keeps the data self-describing
             "interface": {
                 "description": tool.get("description", f"MCP tool from {server_name}"),
                 "inputs": [],  # MCP tools don't read from shared store, only from params
--- a/src/pflow/mcp/types.py
+++ b/src/pflow/mcp/types.py
@@ -75,15 +75,6 @@ class InterfaceSchema(TypedDict, total=False):
     mcp_metadata: dict[str, Any]
 
 
-class RegistryEntry(TypedDict):
-    """Registry entry for MCP nodes."""
-
-    class_name: str
-    module: str
-    file_path: str
-    interface: InterfaceSchema
-
-
 # JSON Schema types
 JSONSchemaType = str | list[str]  # Can be string or union type like ["string", "null"]
 JSONSchemaValue = None | bool | int | float | str | list | dict[str, Any]
--- a/src/pflow/registry/registry.py
+++ b/src/pflow/registry/registry.py
@@ -462,9 +462,23 @@ class Registry:
             return False
 
     def _refresh_core_nodes(self, nodes: dict[str, dict[str, Any]]) -> dict[str, dict[str, Any]]:
-        """Refresh core nodes while preserving user and MCP nodes."""
-        # Preserve non-core nodes
-        preserved = {name: data for name, data in nodes.items() if data.get("type") in ("user", "mcp")}
+        """Refresh core nodes while preserving every non-core node.
+
+        The refresh re-discovers core nodes, so everything that isn't core (user,
+        mcp, or an untyped legacy entry) must survive. Keyed on "not core" rather
+        than an allowlist of known types so it fails safe: an entry we don't
+        recognize is kept, never silently dropped. This also self-heals MCP
+        entries synced before they were stamped with type="mcp" — their absent
+        type is not "core", so they're preserved.
+
+        Freshly-discovered core metadata wins on a name collision: preserved
+        entries are merged *under* the fresh scan, not over it. So a stale entry
+        sharing a core node's name (e.g. an untyped "shell" left by the removed
+        `registry scan` CLI) can neither resurrect nor shadow a core node — the
+        broadened predicate can only add non-core names, never override core ones.
+        """
+        # Preserve every non-core node (see docstring: fail-safe, self-healing).
+        preserved = {name: data for name, data in nodes.items() if data.get("type") != "core"}
 
         # Re-discover core nodes — _auto_discover_core_nodes stamps
         # last_core_scan with a PRE-scan timestamp (race-safe).
@@ -473,8 +487,11 @@ class Registry:
         # Reload — populates self._registry_last_scan from the pre-scan stamp.
         refreshed = self._load_from_file()
 
-        # Merge preserved nodes back in
-        refreshed.update(preserved)
+        # Merge preserved nodes UNDER the fresh core scan: setdefault adds a
+        # preserved entry only when its name is free, so fresh core metadata
+        # always wins a name collision (see docstring).
+        for name, data in preserved.items():
+            refreshed.setdefault(name, data)
 
         # Save the merged result, propagating the pre-scan timestamp so the
         # race-safety is preserved through the final merge write. Without this,
--- a/tests/test_mcp/test_mcp_discovery_critical.py
+++ b/tests/test_mcp/test_mcp_discovery_critical.py
@@ -141,6 +141,8 @@ class TestMCPRegistrarCritical:
             assert entry["class_name"] == "MCPNode"
             assert entry["module"] == "pflow.nodes.mcp.node"
             assert entry["file_path"] == "virtual://mcp"
+            # type="mcp" so the entry survives core-node refresh (issue #462)
+            assert entry["type"] == "mcp"
 
             # Interface MUST have these exact fields
             interface = entry["interface"]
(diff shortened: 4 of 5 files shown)


spinje:
**Codex (C1 — untyped entry shadowing a fresh core node):** Addressed, but via a different fix than the suggested sentinel revert. The mechanism is real (`refreshed.update(preserved)` let a preserved entry overlay fresh core metadata), but no *current* path can create an untyped core-named entry — `_auto_discover_core_nodes` always stamps `type="core"`, and the only untyped writer (`update_from_scanner`) lost its sole caller when `pflow registry scan` was removed in Task 151. It can only exist in a legacy on-disk registry from that removed CLI. Rather than revert to an `mcp-*`/sentinel allowlist (which loses the fail-safe property), I closed the gap by construction: preserved entries are now merged **under** the fresh scan (`setdefault`, not `update`), so fresh core metadata always wins a name collision. Added `test_refresh_does_not_shadow_core_node_with_untyped_entry` to lock it.

**Gemini (G1 — `isinstance(data, dict)` guard):** Declined as proposed. There are zero such guards anywhere in `registry.py`; the `load()` filtering loop calls `node_data.get(...)` unguarded on every load, so a malformed `{"foo": null}` entry crashes there *before* `_refresh_core_nodes` is ever reached. Guarding only this line creates false robustness, and this PR doesn't introduce the fragility (the prior predicate had the identical `.get()`). The right place for a fix, if wanted, is a one-time non-dict filter in `_load_from_file` — tracked as separate hardening, not this PR.

**Claude review (S1/S2/S3):**
- S1 (add `type` to `RegistryEntry`): the TypedDict was unused dead code, so I **deleted** it rather than expand an inaccurate, unreferenced type.
- S2 (docstring note on shadowing): done — the `_refresh_core_nodes` docstring now explains that fresh core wins on collision and why a stale entry can't resurrect a core node.
- S3 (assert legacy entry survives intact): done — the preservation test now asserts the untyped MCP entry is byte-for-byte unchanged through the merge, not just present.

`make check` clean; full suite 7822 passing.

PR #548, opened 2026-07-01
fix: release the browser's 6-connection cap for multi-tab live-overlay viewers (#539)

### LINKED ISSUE #539: Live overlay: SSE over HTTP/1.1 exhausts the browser's 6-connection-per-origin cap with multiple graph tabs
## Problem

Opening several live-overlay **graph** tabs at once eventually makes new tabs hang on "loading" — the page never paints. Root cause (all confirmed against the running server):

1. **Browsers cap concurrent connections at 6 per origin** (Chrome & Firefox; an HTTP/1.1 client-side limit).
2. **Each graph tab holds a *persistent* SSE connection** — `new EventSource("/api/events?…")` (`web/src/api/events.ts`) stays open for the life of the tab to receive live run events (`node.start`, completions, `run-complete`, `run-stopped`, `run-stale`).
3. **The server speaks HTTP/1.1** (uvicorn over plain localhost; `curl -sI` → `HTTP/1.1 200 OK`), so the 6-per-origin cap applies.

So once ~6 graph tabs each hold their SSE open, the browser has **zero free connection slots**, and a new tab's startup fetches (`/api/graph`, `/api/source`, `/api/version`, and its own `/api/events`) queue behind the held-open streams that never close → the tab "just loads" forever. This is the classic SSE-over-HTTP/1.1 limitation (MDN warns about exactly this with multiple tabs).

> Note: only **graph** tabs consume a slot (they hold the SSE). **Catalog** tabs do short-lived fetches that complete and free their slot — they don't count.

## Proposed fix (preferred): close the SSE on a hidden tab, reopen on focus

Release the connection slot for backgrounded tabs (`visibilitychange → hidden` ⇒ close the `EventSource`; `→ visible` ⇒ reopen). A slot is then held only per **visible** graph tab, and you rarely have >6 *visible*. Backgrounded tabs pause live updates (fine — you're not watching them) and catch up via the existing `snapshot()` on reopen.

Alternatives (heavier, likely overkill for a local single-user viewer):
- **HTTP/2** — the real fix (multiplexes ~100s of streams over one socket, cap vanishes), but needs TLS/h2c on a localhost dev server.
- **SharedWorker** — one SSE shared across all tabs, fanned out. The "proper" multi-tab solution, but complex.
- **Accept as a v1 boundary** — it's a local single-user tool; this is the current behavior.

## ⚠ CRITICAL — do NOT regress these prior fixes

A close-on-hidden mitigation **introduces a visibility-driven connection lifecycle**, which the current design *deliberately avoids*. The implementer MUST make it coexist with — not replace or fight — the existing machinery:

1. **The reconnect is `onerror`-driven and trigger-agnostic — keep it that way.** `events.ts` recovers from server-restart / sleep-wake / network-blip / tab-freeze uniformly via `onerror` → close + fixed-delay reopen, **without ever reading `readyState` or `visibilityState`**. `web/src/hooks/CLAUDE.md` states the invariant explicitly: *"SSE connection recovery (`api/events.ts`) deliberately must NOT [key on `visibilitychange`]"* (only the camera re-frame legitimately does). Close-on-hidden adds a visibility trigger for the connection — it must layer on top of, not supplant, the `onerror` path.
2. **Preserve the single-flight invariant:** at most ONE `EventSource` live and at most ONE reconnect pending at any time. The `stopped` flag, the `retry === null` guard, and close-before-schedule are load-bearing. A visibility-close racing an `onerror` reconnect (or a reopen-on-visible racing a pending retry timer) must NOT produce two live sources or resurrect a torn-down one.
3. **Don't reintroduce the "backgrounded tab stuck in CONNECTING" bug.** Native `EventSource` auto-reconnect is unreliable for backgrounded/frozen tabs — the whole reason the manual `onerror` reconnect exists. Reopening on `→ visible` must not assume a prompt connect while the tab is still throttled; closing on `→ hidden` must tear down cleanly (no half-open lingering).
4. **Server-side tailer ref-counting + keepalive-linger (`src/pflow/ui/server.py`).** Closing on hidden fires the server's disconnect cleanup (`unregister` + `release_tailer`); frequent hide/show churns register/release. Keep intact: (a) `release_tailer` stops a `(workflow_key, run_id)` tailer ONLY when its LAST viewer leaves (ref-counted); (b) `ensure_tailer` treats a DONE task as ABSENT and replaces it — the **dead-pinned-tailer-reuse Critical** fix from the Phase-1+2 deep-review (silent blank canvas); (c) the ~15s keepalive-linger window where a dropped conn lingers in `_conns` until its next failed write (`windows` can transiently over-count by 1 — benign, self-corrects). A reopen within that window must stay benign.
5. **Visibility REPORTING must keep working.** `reportVisibility` → `/api/visibility` is separate from the connection lifecycle. On reopen a NEW `conn_id` is issued, so reporting must re-fire on the new connection (the current code "corrects every newly registered connection immediately"). Reconcile the accounting: a closed-because-hidden conn is *gone*, not "hidden" — the Point `windows`/visibility counts must stay coherent.
6. **The hidden-tab camera re-frame (`useCameraNavigation`) is a SEPARATE visibility concern** that legitimately re-fits on `visibilitychange → visible` for a change-while-hidden. The new SSE listener must not interfere with or duplicate it.
7. **Run-state catch-up on reopen.** A backgrounded live-run tab pauses receiving deltas; on reopen it re-subscribes and must catch up via `snapshot()` (node states + the latched `stopped` / `stale_version`). Verify a run that COMPLETED or went STALE while hidden still surfaces correctly on return (incl. the known empty-first-snapshot catch-up for a pinned replay).
8. **Ctrl+C clean shutdown** (`_SHUTDOWN_SENTINEL` + `timeout_graceful_shutdown` + `handle_exit → _Hub.shutdown`) exists because long-lived SSE don't drain. Fewer open SSE only helps, but re-verify shutdown stays clean.

## Verification

- Open 8+ live graph tabs; backgrounded ones release their slot → the 7th+ visible tab loads without stalling.
- Background a live-run graph tab, refocus → it catches up via snapshot and resumes deltas; a run that finished/went stale while hidden surfaces correctly.
- Rapidly hide/show a tab → no duplicate `EventSource`, no orphaned tailer, no console errors, server `windows` count coherent.
- Server restart / sleep-wake / network blip while a tab is VISIBLE → still recovers via `onerror` (existing path unbroken).
- Ctrl+C still shuts the server down cleanly.

## References

- `web/src/api/events.ts` — `subscribe()` (the `onerror` reconnect, single-flight guard, `reportVisibility`).
- `web/src/hooks/CLAUDE.md` — the "SSE connection recovery must NOT key on `visibilitychange`" invariant + the camera re-frame exception.
- `src/pflow/ui/server.py` — `_Hub` / `ensure_tailer` / `release_tailer` / `windows_for_run`, the keepalive-linger, `_Hub.shutdown`.
- Task 173 progress log (`.taskmaster/tasks/task_173/implementation/progress-log.md`): the "`pflow ui` Ctrl+C" entry (long-lived SSE), the Phase-1+2 deep-review "dead pinned tailer reused" Critical, the tailer-scan perf entries.

Surfaced while reviewing the live overlay (Task 173): multiple open graph tabs stalled new-tab loads. Not a blocker for the single-tab common case.

spinje:
## Summary

`pflow ui` graph tabs each held a **persistent SSE** for the tab's whole life. Browsers cap ~6 concurrent connections per origin over HTTP/1.1, so once ~6 graph tabs each held their SSE open, a new tab's startup fetches queued behind the held-open streams forever and the tab hung on "loading".

This closes the SSE when a graph tab is **hidden** (freeing its connection slot) and reopens it when **shown**, so a slot is held only per *visible* graph tab. A reopened/new tab catches up via server-side **latches** replayed on (re)subscribe — for the run overlay, the agent's current **Point** (focus/frame/clear), **and** the agent's current **run selection** (`select-run`), so the agent can steer an open window whether it's visible or backgrounded.

Fixes #539

## Changes

- **Cap fix (`web/src/api/events.ts`)** — refactor `subscribe()` into `dropSource`/`close`/`open`; `open()` is the single connection chokepoint (guarded on `source`/`stopped`/hidden) that closes on `visibilitychange → hidden` and reopens on visible. `onerror` recovery stays trigger-agnostic (never reads `visibilityState`); only `open()`'s gate does. Single-flight (one live `EventSource`, one pending retry) holds across every hide/show/error/unsubscribe interleave.
- **Point latch (`src/pflow/ui/server.py`)** — `focus`/`frame`/`clear` carry a monotonic `epoch`; the latest is remembered per workflow (`_Hub.set_point`/`point_for`) and replayed to any (re)subscribing Viewer after the run snapshot, so a Point sent while a tab was hidden still lands on return.
- **Steer an open window to a run (`select-run` latch)** — `select-run` is latched the same way (`set_run`/`run_for`, its own store, sharing the epoch counter), so the agent can steer a window to a run whether it's **visible** (delivered live) or **backgrounded** (caught up on return). This also removes a silent-failure: `pflow ui X --run Y`'s reuse guard can't tell a live tab from a just-closing one, but the latch now delivers on reopen regardless, so the message is honest again (`steering it to run …`).
- **Epoch dedup + restart fence (`events.ts`)** — the client applies a latched command only if its `epoch` is newer than the last it showed **on that channel** (`point` vs `run` are independent), so a returning/new tab catches up without clobbering the user's own navigation; a per-process `boot_id` on the `connected` frame re-bases the epochs across a server restart.
- **Docs** — `web/src/hooks/CLAUDE.md` (recovery-vs-presence invariant), `src/pflow/ui/CLAUDE.md` (latch + windows semantics), `src/pflow/guide/features/ui.md` (agent-facing note).

## Explanation

The reconnect machinery Task 173 built is deliberately `onerror`-driven and must NOT key on `visibilitychange` — so the fix separates two concerns that share one single-flight state machine: connection **recovery** (still `onerror`-driven, trigger-agnostic) and connection **presence** (new: visibility-gated in `open()`). Run-overlay state is fully reopen-safe — a fresh `RunTailer` reconstructs the run from the trace, and `snapshot()` carries node statuses + the `stopped`/`stale_version` latches. Task 175's run launch is a detached subprocess, so closing a tab can't kill a run.

The epoch is the minimal mechanism that lets the client distinguish "a latch I already applied" (skip) from "an agent command issued while I was hidden" (apply) without the server tracking per-connection delivery; `boot_id` fences it against a fresh process restarting the counter at 1. Latching `select-run` on a separate channel makes the latch — not the laggy connection count — the delivery guarantee for steering.

## Follow-up

- **#549** — remove the now-vestigial per-connection visibility machinery (`/api/visibility`, `set_visibility`, `_Conn.visibility`, the `?visibility=` param, the per-window `visibility` field). With close-on-hidden + latching, a live connection is always visible, so the whole apparatus can only ever report `"visible"`. Kept out of this PR because it's a wire-contract change.

## Testing

Server (`tests/test_cli/test_ui_interaction_server.py`): hub latch + monotonic epoch (`set_point`/`set_run`, separate stores, shared sequence); `boot_id` restart fence; focus/clear/select-run latch with no live windows; clear supersedes focus; zero/ambiguous match mints no epoch; `events()` replays the latched **Point** and the latched **run selection** to a new connection (frame order `connected → snapshot → point → run`). CLI (`test_ui_commands.py`): `--run` reuse steers via `select-run` and reports the steer.

Frontend (`web/src/api/events.test.ts`): starts-hidden→no-connect; hidden→close/visible→reopen; rapid hide/show→single source; retry-cancel-on-hidden (`vi.getTimerCount() === 0`); epoch dedup (newer applies / stale + already-shown skipped across a reopen); `select-run` dedup on its own channel + point/run channel independence; `boot_id` reset; absent-epoch back-compat. The `FakeEventSource` double was hardened to model WHATWG closed-source event suppression.

Suites: **web 693 passed** (tsc clean); **Python 8283 passed**; `make check` (ruff + mypy 239 files + deptry) clean.

Manual verification (real browser + real HTTP against a running server):
- hide→show reopened **exactly one** EventSource (`openedAfterToggle: 1`); a `focus` latched while hidden **panned the camera** to an off-screen node on return; a live `focus` to a visible tab delivered instantly.
- steering a workflow to a run with **no viewer connected** latched it, and a fresh viewer caught up (`connected → run-snapshot → select-run`).
- `windows` drops to 0 after a tab is backgrounded (the slot released); the `connected` frame carries `boot_id`.

Two `/deep-review` passes (plan + code: concurrency, silent-failures, feature-interactions, test-fidelity, simplicity): 0 confirmed Critical/Warning; a flagged camera-pan regression was empirically refuted; suggestions applied. An external review's findings were addressed (honest `--run` message → now a confident steer via the latch; latch-invariant code comments).

### DIFF
--- a/src/pflow/cli/commands/ui.py
+++ b/src/pflow/cli/commands/ui.py
@@ -456,7 +456,15 @@ def serve_cmd(
                 health = _probe_health(port, reuse_workflow)
                 if health is not None and _int_field(health, "windows") > 0:
                     _point_request(ctx, port, reuse_workflow, "select-run", run, output_json=False)
-                    click.echo(f"pflow UI already showing {reuse_workflow} — switched it to run {run}", err=True)
+                    # `windows > 0` can't tell a live-visible tab from a just-backgrounded tab's ≤15s
+                    # lingering-closed connection, and `select-run` isn't latched (Issue #539) — so report
+                    # honestly rather than assert success; an agent can re-run to open a fresh tab if nothing
+                    # changed. (A future fix could open a pinned tab here instead of trusting `windows`.)
+                    click.echo(
+                        f"pflow UI already showing {reuse_workflow} — asked it to switch to run {run}. "
+                        f"If the view didn't change, that tab was closing; re-run to open a fresh pinned tab.",
+                        err=True,
+                    )
                     ctx.exit(0)
                     return
             url = _serve_url(port, reuse_workflow, no_auto_update, run=run)
--- a/src/pflow/cli/commands/ui.py
+++ b/src/pflow/cli/commands/ui.py
@@ -456,15 +456,10 @@ def serve_cmd(
                 health = _probe_health(port, reuse_workflow)
                 if health is not None and _int_field(health, "windows") > 0:
                     _point_request(ctx, port, reuse_workflow, "select-run", run, output_json=False)
-                    # `windows > 0` can't tell a live-visible tab from a just-backgrounded tab's ≤15s
-                    # lingering-closed connection, and `select-run` isn't latched (Issue #539) — so report
-                    # honestly rather than assert success; an agent can re-run to open a fresh tab if nothing
-                    # changed. (A future fix could open a pinned tab here instead of trusting `windows`.)
-                    click.echo(
-                        f"pflow UI already showing {reuse_workflow} — asked it to switch to run {run}. "
-                        f"If the view didn't change, that tab was closing; re-run to open a fresh pinned tab.",
-                        err=True,
-                    )
+                    # `select-run` is latched (Issue #539), so this steers the open window whether it's
+                    # visible (live) or backgrounded (caught up on return) — even if `windows` was counting a
+                    # just-closing tab, the latch still delivers on reopen. So report the steer plainly.
+                    click.echo(f"pflow UI already showing {reuse_workflow} — steering it to run {run}", err=True)
                     ctx.exit(0)
                     return
             url = _serve_url(port, reuse_workflow, no_auto_update, run=run)
--- a/src/pflow/guide/features/ui.md
+++ b/src/pflow/guide/features/ui.md
@@ -44,6 +44,9 @@ windows and Watch what they deliberately click:
 - `pflow ui clear-focus <workflow>` — clear the current focus.
 - `pflow ui user-activity [workflow]` — read recent clicks and view changes.
 
+`focus`/`frame`/`clear-focus` reach every matching Viewer — including one whose tab is
+currently backgrounded; the highlight is applied when the user returns to it.
+
 **A target is just the name you already read in the `.pflow.md` — there is no
 separate notation to learn:**
 
--- a/src/pflow/ui/CLAUDE.md
+++ b/src/pflow/ui/CLAUDE.md
@@ -175,13 +175,26 @@ is `run_node.read_run_inputs`; sync handler, threadpooled, touches no hub state)
   run), plus the Task-173 run/overlay stream (`run-snapshot`/`run-events`/
   `run-complete`/`run-reset`/`run-not-found`/`run-stopped`/`run-stale`). (An earlier
   note here claimed the envelope "defines no run/trace event schema" — Task 173
-  invalidated that; the run-* schema is the overlay's, see `run_tailer`.)
+  invalidated that; the run-* schema is the overlay's, see `run_tailer`.) The `connected`
+  frame also carries a per-process **`boot_id`** (Issue #539 restart-fence for the
+  Point-epoch dedup). Connection **presence** is now visibility-driven: the frontend
+  CLOSES this stream when the tab is hidden (freeing one of the browser's ~6 per-origin
+  HTTP/1.1 connection slots) and reopens on show, so a graph tab holds a slot only while
+  visible. On every (re)subscribe the server replays `snapshot()` + the latched Point
+  (below) to catch the tab up.
 - `POST /api/command` validates a `focus`/`frame` target against a fresh graph, or
   broadcasts `clear` / `select-run` (both pass-through — `clear` before the target is
   read, `select-run` after, carrying the run id in `target` with NO graph resolution;
   a stale run id surfaces the frontend's `run-not-found`, never a server error). It
   reports `sent_to`, per-window visibility, and the canonical `workflow_key` — messages
-  queued to live connections, not browser apply acknowledgments.
+  queued to live connections, not browser apply acknowledgments. **Issue #539:** each
+  `focus`/`frame`/`clear` also carries a monotonic **`epoch`** and is **latched** per
+  `workflow_key` (the latest replaces the prior), which `/api/events` replays after the
+  snapshot so a tab that was hidden when the command fired — or a brand-new tab — catches
+  up to the agent's current highlight; the client dedups by `epoch` (re-baselined on a
+  `boot_id` change) so the replay is idempotent. `select-run` is **not** latched — each
+  tab keeps its own `?run=` pin — so a `select-run` to a hidden tab is dropped (it reaches
+  a visible tab, or `--run` opens one).
 - `POST /api/interaction` records deliberate user actions; `GET /api/activity`
   returns the bounded, newest-first snapshot. `POST /api/visibility` updates one
   connection. All POSTs require `application/json`.
@@ -190,11 +203,19 @@ is `run_node.read_run_inputs`; sync handler, threadpooled, touches no hub state)
   resolvable `workflow` is supplied (`windows = len(windows_for(key))`, **no graph
   build**). An unresolvable workflow reports identity only (no 404 — a liveness probe
   must answer regardless, unlike `events()`/`command()`). It reads the hub, so it is
-  `async def` per the invariant below. **`windows` can transiently over-count by 1**
-  for up to one `_KEEPALIVE_S` cycle after a Viewer's `onerror` reconnect to *this*
-  server (the dropped connection lingers until its next keepalive write fails); a
-  server *restart* frees the whole hub, so that path is clean. Benign for a local
-  single-user viewer; the count self-corrects.
+  `async def` per the invariant below. **`windows` can transiently over-count**
+  for up to one `_KEEPALIVE_S` cycle after a Viewer reconnects to *this* server (the
+  dropped connection lingers until its next keepalive write fails); a server *restart*
+  frees the whole hub, so that path is clean. Since Issue #539 a hidden tab CLOSES its
+  stream (rather than reporting `visibility:hidden`), so `windows` also **drops** a
+  backgrounded tab after its ≤`_KEEPALIVE_S` linger, `_dispatch_report`'s per-window
+  `visibility` no longer surfaces a `hidden` window (a live conn is always a visible tab),
+  and rapid hide/show can over-count by more than 1 (several lingering-closed conns
+  coexist). All benign for a local single-user viewer; the count self-corrects. **Caveat
+  (`pflow ui X --run Y`):** the reuse guard keys on `windows>0`, which can't tell a
+  live-visible tab from a lingering-closed one — if the sole tab was hidden <`_KEEPALIVE_S`
+  ago it POSTs `select-run` (unlatched) to a dead connection and exits without opening a
+  tab; a re-run after the linger fixes it. (Known ≤15 s limitation.)
 
 Name-opened and path-opened Viewers share one resolved workflow key. Point targets
 cross the wire only as structural refs; never send positional flat ids between
--- a/src/pflow/ui/server.py
+++ b/src/pflow/ui/server.py
@@ -135,6 +135,12 @@ class _Hub:
         # `_conns`/tailers): a brand-new tab minutes after the last one closed must still catch up, so pruning
         # on last-window-close would defeat the feature — bounded by O(#workflows ever pointed at), negligible.
         self._points: dict[str, dict[str, object]] = {}
+        # The agent's current run selection per workflow_key: `select-run` is latched the same way, so the
+        # agent can STEER an open window to a run whether it's visible (delivered live) or backgrounded
+        # (caught up on return). Separate from `_points` because focus and run selection are orthogonal
+        # state — a returning tab catches up to BOTH. Shares `_point_epoch`; the client dedups it on a
+        # SEPARATE baseline. Same never-evicted rationale as `_points`.
+        self._selected_run: dict[str, dict[str, object]] = {}
         self._point_epoch = itertools.count(1)
         self.boot_id = uuid.uuid4().hex
         # Task 173: one live-run tailer per (workflow_key, run_id) — started on first viewer subscribe and
@@ -182,23 +188,42 @@ class _Hub:
         run-events. A subset of ``windows_for`` (which spans every run of the workflow, for Point)."""
         return [c for c in self._conns.values() if c.workflow_key == workflow_key and c.run_id == run_id]
 
-    def set_point(self, workflow_key: str, message: dict[str, object]) -> dict[str, object]:
-        """Latch the agent's current Point for a workflow and return the broadcast envelope with its epoch.
+    def _stamp(self, message: dict[str, object]) -> dict[str, object]:
+        """A fresh, immutable-once-minted envelope carrying the next monotonic Point epoch.
 
-        Issue #539: focus/frame/clear are transient broadcasts with no snapshot replay, so a tab whose SSE
-        was closed while hidden would miss them. We stamp each with a monotonic ``epoch`` and remember the
-        latest per workflow_key; ``events()`` replays it to every new/reconnecting Viewer, and the client's
-        epoch-dedup applies it only if newer than what it already showed — a returning tab catches up without
-        clobbering the user's own navigation. The epoch encodes BROADCAST order, not request-arrival order
-        (``clear`` stamps before focus/frame's off-loop build), which matches what live clients already see.
-
-        The returned envelope is stored in ``_points`` AND broadcast to every conn as the SAME dict — it is
-        immutable once minted (this method always builds a fresh dict; no consumer mutates it, only
-        ``json.dumps``), so sharing the reference between the latch and the broadcast queues is torn-read-free.
+        Always a NEW dict (the store and the broadcast queues share this one reference; no consumer mutates
+        it, only ``json.dumps``), so cross-queue sharing is torn-read-free. The epoch encodes BROADCAST order,
+        not request-arrival order (``clear`` stamps before focus/frame's off-loop build) — what live clients
+        already see. Shared by both latches so focus and run selection stay on one monotonic sequence.
         """
-        envelope = {**message, "epoch": next(self._point_epoch)}
+        return {**message, "epoch": next(self._point_epoch)}
+
+    def set_point(self, workflow_key: str, message: dict[str, object]) -> dict[str, object]:
+        """Latch the agent's current Point (focus/frame/clear) and return the epoch-stamped broadcast envelope.
+
+        Issue #539: these are transient broadcasts with no snapshot replay, so a tab whose SSE was closed
+        while hidden would miss them. We remember the latest per workflow_key; ``events()`` replays it to every
+        new/reconnecting Viewer, and the client's epoch-dedup applies it only if newer than what it already
+        showed — a returning tab catches up without clobbering the user's own navigation.
+        """
+        envelope = self._stamp(message)
         self._points[workflow_key] = envelope
         return envelope
+
+    def set_run(self, workflow_key: str, message: dict[str, object]) -> dict[str, object]:
+        """Latch the agent's current run selection (``select-run``) and return the epoch-stamped envelope.
+
+        Mirrors ``set_point`` on a separate store so the agent can STEER an open window to a run: a visible
+        window gets it live, a backgrounded one catches up on return via the ``events()`` replay. The
+        client dedups it on its own baseline, so re-steering works and a user's later manual run pick isn't
+        clobbered by a replay of an already-applied selection.
+        """
+        envelope = self._stamp(message)
+        self._selected_run[workflow_key] = envelope
+        return envelope
+
+    def run_for(self, workflow_key: str) -> dict[str, object] | None:
+        return self._selected_run.get(workflow_key)
 
     def point_for(self, workflow_key: str) -> dict[str, object] | None:
         return self._points.get(workflow_key)
@@ -575,10 +600,11 @@ async def events(request: Request) -> Response:
             # it. The epoch-carrying envelope is deduped client-side, so replaying it is idempotent. Unlike
             # the snapshot above there is NO follow-up stream to re-sync this, but the trigger for a dropped
             # replay is near-impossible (the fresh 64-slot queue holds only the snapshot at this point).
-            latched = hub.point_for(workflow_key)
-            if latched is not None:
-                with contextlib.suppress(asyncio.QueueFull):
-                    conn.queue.put_nowait(latched)
+            # ...and to the agent's current run selection, so `select-run` steers a returning/new window too.
+            for latched in (hub.point_for(workflow_key), hub.run_for(workflow_key)):
+                if latched is not None:
+                    with contextlib.suppress(asyncio.QueueFull):
+                        conn.queue.put_nowait(latched)
             while conn.active:
                 try:
                     message = await asyncio.wait_for(conn.queue.get(), timeout=_KEEPALIVE_S)
@@ -665,10 +691,11 @@ async def command(request: Request) -> Response:
     # skips resolve_target entirely (a run isn't a node/edge). The browser's selectRun applies it (honoring
     # its own re-pick guard); a stale/unknown id surfaces the frontend's run-not-found path in the open
     # Viewer, never a server error — so no server-side run validation. Placed AFTER `target` is read (unlike
-    # `clear`, which returns before it) since the id lives in `target`.
+    # `clear`, which returns before it) since the id lives in `target`. Issue #539: latched (like Point) so
+    # the agent can steer a backgrounded/returning window to the run, not just a currently-visible one.
     if command_type == "select-run":
-        conns = hub.broadcast(workflow_key, {"type": "select-run", "run": target})
-        return _json(_dispatch_report(workflow_key, conns))
+        steered = hub.set_run(workflow_key, {"type": "select-run", "run": target})
+        return _json(_dispatch_report(workflow_key, hub.broadcast(workflow_key, steered)))
 
     try:
         # Validation/build can recurse through large nested workflows. Keep it
--- a/src/pflow/ui/server.py
+++ b/src/pflow/ui/server.py
@@ -131,7 +131,9 @@ class _Hub:
         # highlight. `boot_id` fences the epoch across a server restart: a fresh process restarts the counter
         # at 1, so the client resets its `lastAppliedEpoch` baseline when this nonce changes on reconnect,
         # else a surviving tab would silently skip the new server's lower-numbered Points. Loop-owned like the
-        # rest of the hub (mutated only from command(), on the event loop).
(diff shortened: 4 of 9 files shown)


spinje:
## Implementation plan

The plan approved (in plan mode) before coding, kept here as the design record. It captures the reasoning, the deep-review findings folded in, and the rejected alternatives.

**Two things evolved after this plan — visible in the commit history:**

- **`select-run` is now _also_ latched** (this plan explicitly *excluded* it). That reversal is what lets the agent **steer an open window to a run** whether it's visible (delivered live) or backgrounded (caught up on return) — and it turns the `pflow ui X --run Y` "≤15 s reuse-guard gap" documented below from a known limitation into a **resolved** one: the latch is the delivery guarantee, so the laggy `windows` count no longer matters.
- **The vestigial visibility-machinery cleanup** (the dead CLI branch + the `_dispatch_report` per-window `visibility` notes below) was split into a dedicated follow-up: #549.

---

# Issue #539 — Release the browser's 6-connection cap for multi-tab live-overlay viewers

## Context

`pflow ui` graph tabs each hold a **persistent SSE** (`EventSource("/api/events?…")`, `web/src/api/events.ts`)
open for the tab's whole life, to receive live run events + agent Point commands. Browsers cap **~6
concurrent connections per origin** over HTTP/1.1 (uvicorn serves plain HTTP/1.1 on localhost). So once
~6 graph tabs each hold their SSE open, a new tab's startup fetches (`/api/graph`, `/api/source`,
`/api/version`, its own `/api/events`) queue behind the held-open streams forever → the tab "just loads."
This is the classic SSE-over-HTTP/1.1 multi-tab exhaustion. Catalog tabs do short-lived fetches that
free their slot; only **graph** tabs (which hold the SSE) count.

**Chosen fix:** close the SSE on a **hidden** tab and reopen it on **visible**. In one browser window only
the *active* tab is `visible`; the rest are `hidden` → hold **0** slots. This fully fixes the reported
single-window-many-tabs case, is frontend-only, ~small, and trivially reversible. It also *reduces server
load* (a hidden tab's 4 Hz trace-tailer is torn down via `release_tailer`).

**One consequence + its remedy:** closing the SSE means an agent `focus`/`frame`/`clear`/`select-run`
broadcast sent *while a tab is hidden* is dropped (these are transient — no snapshot replay). We keep
delivery for `focus`/`frame`/`clear` via a small **server-side Point latch** replayed on (re)subscribe,
epoch-guarded so it catches up a returning/new tab to the agent's *current* highlight **without clobbering**
the user's own navigation. (`select-run` is excluded — it conflicts with each tab's `?run=` pin.)

Why NOT the alternatives (all heavier, decided with the user): **SharedWorker** — a shared socket doesn't
release slots (it stays open for any tab) and only helps if it *multiplexes* many `(workflow,run)`
subscriptions over one socket, which needs a **server protocol redesign** + re-homing the reconnect
machinery into a jsdom-untestable worker. **HTTP/2** — the real fix, but browsers require TLS for h2, which
is friction for a `127.0.0.1` tool. Both are overkill for a local single-user viewer.

Verified (via full reads of the Task 172/173/175 progress logs + the server code): run-overlay state is
**fully reopen-safe** — a fresh `RunTailer` reads the trace from byte 0 (per-event `file.flush()`), and
`snapshot()` carries every node status + the `stopped`/`stale_version` latches, so a reopened tab (a new
first-subscriber to a possibly-fresh tailer) catches up completely. `ensure_tailer` already treats a DONE
task as absent and replaces it (the "dead pinned tailer" fix), `release_tailer` is ref-counted, and the
~15 s keepalive-linger over-count is documented-benign. Task 175's run launch is a **detached subprocess**
independent of the SSE, so closing a tab can't kill a run. **The server needs no changes for the cap fix**;
the only server change is the additive Point latch.

## Approach

Three coordinated pieces. The cap fix (A) is the core; the latch (B) + epoch dedup (C) preserve agent
delivery.

### A. Cap fix — visibility-driven connection presence (`web/src/api/events.ts`)

Refactor `subscribe()` into **three** small helpers so a single chokepoint owns connection creation, and
layer a visibility trigger on top of the **untouched** `onerror` recovery. **Exact shape (the two CRITICAL
bugs a naive refactor introduces — a permanent reconnect deadlock — are called out inline):**

```
const dropSource = () => { source?.close(); source = null; connId = null; };   // MUST null source (see below)

const close = () => { if (retry) clearTimeout(retry); retry = null; dropSource(); };  // hidden path + unsubscribe

const open = () => {
  if (stopped || source !== null || visibility() === "hidden") return;   // single chokepoint + visibility gate
  const params = ...; if (runId) params.set("run", runId);               // ?run= pin preserved across reopen
  source = new EventSource(`/api/events?${params}`);
  source.onmessage = (…) => { … epoch dedup (C) … };                     // unchanged except (C)
  source.onerror = () => {                                               // stays trigger-agnostic — never reads visibilityState
    dropSource();                                                        // ← P1: MUST null source, or open()'s `source!==null` gate deadlocks reconnect forever
    if (!stopped && retry === null)
      retry = setTimeout(() => { retry = null; open(); }, RETRY_MS);     // ← P2: keep the WRAPPER (not bare setTimeout(open)) so the retry===null guard re-arms
  };
};

const onVisibilityChange = () => (visibility() === "hidden" ? close() : open());
```

- Mount: `open()` (no-ops if the tab starts hidden). Listener: `addEventListener("visibilitychange",
  onVisibilityChange)` — this **replaces** the old `reportVisibility` listener. Teardown: `stopped = true;
  removeEventListener(...); close()`.
- **P1/P2 are the load-bearing corrections:** the current `onerror` does *not* null `source` (it relies on
  `connect()` reassigning it) and today's wrapper nulls `retry` before reconnecting. With the new
  `source !== null` gate in `open()`, `onerror` MUST null `source` (via `dropSource`) or the gate returns
  early forever; and it MUST keep the `retry = null; open()` wrapper or the `retry === null` guard never
  re-arms after a second drop (server-restart / repeated sleep-wake). This factoring preserves the current
  churn semantics (retry timed from the *first* drop).
- **The visibility gate lives ONLY in `open()`** — so a queued reconnect can never resurrect a hidden tab,
  and an initial background-tab mount won't connect until first visible. `onerror` never reads
  `visibilityState` (the recovery invariant, `hooks/CLAUDE.md:69-70`).
- **No `visibility:hidden` report on hide** — the conn is *gone*, not "hidden". `reportVisibility()` stays
  fired only from the `connected` message (always reports "visible" now, since a live conn ⇒ a visible tab).
  The server's ~15 s linger over-count is documented-benign (self-corrects on the next failed keepalive).

Single-flight invariant preserved: at most one live `EventSource`, at most one pending `retry`; a
hide-`close()` cancels any pending reconnect; the `open()` gate + `stopped` flag prevent resurrection. The
`useCameraNavigation` hidden-tab re-frame hook keeps its own separate listener and is untouched (its
trigger — a focus change *while hidden* — now essentially never fires, since Points arrive on reopen while
visible; harmless, leave it).

### B. Point latch — replay the agent's current highlight on (re)subscribe (`src/pflow/ui/server.py`)

- `_Hub.__init__` gains `self._points: dict[str, dict[str, object]] = {}` (workflow_key → the
  **epoch-carrying envelope**; the epoch rides *inside* the envelope — no separate tuple),
  `self._point_epoch = itertools.count(1)`, and **`self.boot_id = uuid4().hex`** (a per-process nonce for
  the restart-fence, C1); plus a tiny `point_for(workflow_key)` accessor. All mutations are dict-only and
  loop-owned (`command()` mutates *after* its `asyncio.to_thread` build returns, on the loop; `stream()`
  reads in the async handler) — consistent with the hub's async-only invariant.
- `stream()`'s `connected` frame (`server.py:527`) gains `"boot_id": hub.boot_id` alongside `conn_id`.
- In `command()` (`server.py:591`):
  - **`clear`** branch (`:609`): mint `epoch = next(hub._point_epoch)`; `env = {"type":"clear","epoch":
    epoch}`; store `hub._points[workflow_key] = env`; `hub.broadcast(workflow_key, env)`.
  - **`focus`/`frame`** — mint + store **only inside the matched==1 path** (after the early return at
    `:639-640`, so a zero/ambiguous match neither mints nor latches): `env = {"type": command_type,
    "target": resolution.descriptor, "epoch": epoch}`; store **unconditionally** (`hub._points[...] = env`
    even with zero live conns — that is the whole point); broadcast `env`.
  - **`select-run`** (`:621`): unchanged — no latch, no epoch.
- In `stream()`, right after the tailer-snapshot enqueue (`server.py:536`): `latched =
  hub.point_for(workflow_key); if latched is not None: with contextlib.suppress(QueueFull):
  conn.queue.put_nowait(latched)`. `_points` is unbounded but keyed by workflow_key (a handful), matching
  the hub's other tiny per-key caches — no eviction needed.

### C. Epoch dedup on the client (`web/src/api/events.ts`)

- Track `let lastAppliedEpoch = 0;` and `let serverBootId: string | null = null;` in the `subscribe()`
  closure (persist across the *internal* reopen; reset per fresh `subscribe()`, i.e. per workflow/run
  change).
- **Restart-fence (C1):** in the `connected` handler, before anything else — `if (typeof message.boot_id
  === "string" && message.boot_id !== serverBootId) { serverBootId = message.boot_id; lastAppliedEpoch =
  0; }`. This re-baselines the epoch whenever the server identity changes (restart), on **either** reconnect
  path (onerror *and* hidden→visible), so a surviving tab never silently skips a restarted server's
  Points. A same-server reopen sees an unchanged `boot_id` → no reset → dedup still holds.
- For the `clear` and `focus`/`frame` arms, dedup by epoch **with an absent-epoch escape hatch**, checked
  **after** the existing `isTarget` validation so a malformed target never bumps the epoch:
  ```
  const epoch = typeof message.epoch === "number" ? message.epoch : null;
  if (epoch !== null && epoch <= lastAppliedEpoch) return;   // skip a replayed/stale latch
  handlers[...](…);                                          // apply
  if (epoch !== null) lastAppliedEpoch = epoch;
  ```
  A missing/non-number epoch ⇒ **always apply, never bump** — this keeps every existing `events.test.ts` /
  `runEvents.test.ts` case (which emit `focus`/`frame`/`clear` with no epoch) green and stays back-compatible.
  `select-run` keeps its existing frontend re-pick guard (no epoch).

Why the epoch (confirmed not over-engineered): the internal reopen preserves GraphView's `focus` React
state, so a tab only needs to catch up to point commands with `epoch > lastAppliedEpoch`. A user's manual
focus (while visible) never bumps `lastAppliedEpoch`, so a replayed *older* agent point is skipped (manual
focus preserved), while a *newer* agent point issued while the tab was hidden applies (catch-up). It also
dedups the reopen-vs-concurrent-broadcast race (a `command()` firing between `register` and the
`point_for` read delivers the same epoch twice → the second is skipped). A brand-new tab starts at
`lastAppliedEpoch = 0`, so it adopts the agent's current highlight — also fixing a latent gap today (a new
tab opened after the agent points currently misses it). A stale descriptor that no longer resolves after a
live-reload is a benign client no-op (`applyPoint` already returns on an unresolvable node/edge target;
`pointHandlers.current` is reassigned every render, so a replay resolves against the *current* graph).

### C1 — MUST FIX (Critical, silent-failures review): epoch desync on server restart

`lastAppliedEpoch` persists across the *internal* reopen (by design), but `_point_epoch` **resets to 1 on
every server restart** (a new `_Hub`). Common in the agent dev-loop (Ctrl+C+relaunch, or the documented
`SO_REUSEADDR` fast-relaunch). After a restart, a surviving tab reconnects (via `onerror` *or* via
hidden→visible `open()`) with a stale high-water `lastAppliedEpoch` (say 15) while the new server mints
1,2,3… → the client **silently skips every agent Point** (`1 <= 15`) with `sent_to>0` reported upstream,
until the counter climbs back past 15. Two tabs of the same workflow then diverge with no signal.

**Fix (chosen — keeps the epoch design self-contained in the transport seam, GraphView untouched):** the
per-process `boot_id` on the `connected` envelope (see B) + the client restart-fence (see C). ~2 server +
~3 client lines, covering **both** reconnect paths — a "reset-on-onerror" shortcut was rejected because it
misses the hidden-across-restart→visible path, which never fires `onerror`.

## Files to modify

- `web/src/api/events.ts` — the cap fix (A) + epoch dedup (C). Update the `subscribe()` docstring to split
  **recovery** (onerror, trigger-agnostic) from **presence** (visibility-gated `open()`/`close()`).
- `web/src/api/events.test.ts` — extend: starts-hidden→no-connect; hidden→close/visible→reopen; rapid
  hide/show→single source; retry cancelled on hidden; onerror-still-reconnects-when-visible (keep green);
  `connected→reportVisibility` on reopen; epoch dedup (newer applies, already-seen skipped across a reopen).
  **Harden the `FakeEventSource` double** (concurrency-review W1): `fail()` and `emit()` must early-return
  when `this.closed`, mirroring the WHATWG spec's closed-source `onerror`/message suppression — the
  single-flight safety argument *relies* on a closed source never firing `onerror` again, and the current
  double doesn't model it, so the new hide/show tests would otherwise exercise an impossible interleaving.
- `src/pflow/ui/server.py` — the Point latch (B): `_Hub` fields + accessors, `command()` mint/attach/store,
  `stream()` replay.
- `tests/test_cli/test_ui_interaction_server.py` — **MUST update two exact-dict assertions** now that
  focus/frame/clear envelopes carry `epoch`: `:171-177` (`focus` → `{...,"epoch":1}`) and `:191` (`clear`
  → `{"type":"clear","epoch":1}`). Leave `:75-76` (direct `hub.broadcast`, no `command()` → no epoch) and
  `:207` (`select-run`, excluded) as-is. Add new latch tests here: focus/frame sets the latch; clear
  updates it; a fresh `/api/events` connection receives the latched point right after `connected` +
  snapshot; epoch increments; `select-run` is NOT latched.
- `tests/test_cli/test_ui.py` — preserve `test_hub_shutdown_ends_streams_with_a_sentinel`,
  `test_ensure_tailer_replaces_a_terminated_tailer`, and any exact-`windows` assertions. **P6:** the latch
  replay inserts a frame between `snapshot` and the keepalive loop, so `test_idle_connection_emits_keepalive_frames`
  stays green **only because `_points` is empty** there — never assert `connected→snapshot→keepalive` after
  a `command()` has run.
- `web/src/hooks/CLAUDE.md:67-70` — rewrite the invariant: connection **recovery** stays onerror-driven and
  trigger-agnostic (never keys on `visibilitychange`); connection **presence** now keys on visibility
  (close-on-hidden / reopen-on-visible) to release the 6-per-origin cap. The two are separate concerns.
- `src/pflow/ui/CLAUDE.md` — note (1) the `windows` count now also drops on backgrounding (a hidden tab is
  gone, not "hidden"), and `_dispatch_report`'s per-window `visibility` no longer reports a backgrounded
  window as "hidden" (it lingers "visible" ~15 s, then disappears) — an intentional trade, not just an
  over-count; (2) the new Point-latch behavior on `/api/command` (focus/frame/clear carry `epoch` + are
  latched per workflow) + `/api/events` (replays the latch after the snapshot).

Reuses (no new machinery): the existing single-flight `source`/`retry`/`connId`/`stopped` state, the
`reportVisibility()` on `connected`, `hub.broadcast` + `_dispatch_report`, the `stream()` snapshot-enqueue
pattern, and the `FakeEventSource` test harness.

## Scope boundary (stated in the PR)

Fixes the reported **many-tabs-in-one-window** case fully. Does **not** fix **>6 simultaneously-*visible*
windows** (side-by-side) — that needs HTTP/2 or multiplexing, the accepted v1 boundary the issue names.

## Known trade-offs (documented, accepted)

- Agent `select-run` to a *hidden* tab is **now dropped** (excluded from the latch). This is a **new**
  drop, not pre-existing: today a backgrounded tab keeps its SSE open and *does* receive the `select-run`
  (it switches on refocus). Accepted — a run switch to a tab you're not watching is moot, and a returning
  tab keeps its own `?run=` pin.
- **`pflow ui X --run Y` reuse-guard gap (transient, ≤15 s — review-caught).** The guard is `windows>0`,
  but `windows` counts a just-backgrounded tab's ≤15 s **lingering-closed** connection — so if the sole
  Viewer was hidden <15 s ago, the CLI POSTs `select-run` to that dead connection and `ctx.exit(0)`
  **without** opening a tab → nothing shows run Y until the linger expires (then a re-run opens a fresh
  pinned tab). My earlier "the `windows>0` guard reroutes to a visible tab" claim was wrong: `windows`
  can't distinguish a live-visible tab from a lingering-closed one. **Recommended:** document as a known
  ≤15 s limitation for this PR (a re-run fixes it); a proper fix (open a fresh pinned tab for the unlatched
  `--run` verb rather than trusting `windows>0`) is a small follow-up in `cli/commands/ui.py`, kept out of
  this PR's scope (listed under Optional follow-ups). Say if you'd rather include it now.
- `/api/health` `windows` under-counts a *long*-hidden tab (closed, linger expired) and can transiently
  **over**-count during rapid hide/show (several lingering-closed conns coexist — the "over-count by 1"
  framing understates this; benign, `ensure_tailer` reuse prevents duplicate tailers and `release_tailer`
  still cleans up). `focus --open` readiness keys on window *count* and opens a fresh *visible* tab when a
  highlight has no live recipient — benign, arguably better.
- Re-focusing a **finished** run's tab after >15 s hidden re-triggers the known sub-second
  empty-first-snapshot flash (a fresh tailer); a quick refocus (<15 s) reuses the lingering tailer → no
  flash. Accepted (existing cost, now occasionally repeated).
- The `cli/commands/ui.py:309-314` "background tab — tell the user to switch to it" hint is now **dead**
  (live conns always report `visible`; hidden tabs close). Not a bug (the latch replaces its purpose for
  focus/frame/clear) — note it in the doc pass so a future reader doesn't puzzle over it.

## Verification

1. **Unit (automated):** `cd web && npx vitest run` (events.test.ts + runEvents.test.ts); `uv run pytest
   tests/test_cli/test_ui_interaction_server.py tests/test_cli/test_ui.py` — **the modified exact-dict
   assertions and all new latch tests live in `test_ui_interaction_server.py`, so that file MUST be in the
   command** (or just run `make test`). Capture the baseline first, then diff. `make check` for lint/types.
2. **Build:** `make ui-build` (the server serves the built bundle, not source).
3. **Real-browser round-trip** (the "windows coherent" loop) via the `screenshot-pflow-web-ui` MCP tooling
   + `curl`: open a graph tab → `curl /api/health?workflow=X` shows `windows:1`; drive a real
   `visibilitychange` (CDP `Emulation.setPageVisibilityState hidden`, if the MCP exposes it) → the
   `/api/events` request closes and `windows` drops to 0 within ~15 s; `→ visible` → reopens, `windows:1`.
   If CDP visibility isn't exposed, fall back to a scripted multi-page open + eyeball, and flag the true
   8-tab symptom for a quick human confirm.
4. **Latch check:** with a tab open, POST a `focus` (`pflow ui focus X <target>`); open a SECOND tab →
   it should adopt the agent's highlight on connect (latch replay). Background→refocus the first tab after a
   `focus` change → it catches up to the newer target, and a manual click before hiding is preserved.
5. **Regression guard:** server restart / sleep-wake / network blip while a tab is VISIBLE still recovers
   via `onerror`; Ctrl+C still shuts down cleanly.

## Low-priority notes to fold in as code comments (from review)

- **Epoch encodes broadcast order, not request-arrival order** (`command()` mints `clear` synchronously but
  `focus`/`frame` after the `asyncio.to_thread` build) — only diverges under concurrent un-awaited POSTs
  (the CLI serializes), and matches what live clients already see. One-line comment so a reader doesn't
  assume request order.
- **Latch-replay `QueueFull` suppression has no re-sync path** (unlike the snapshot, which re-syncs via
  streamed deltas). The trigger is near-impossible (fresh 64-slot queue holds ≤1 item at replay time), so a
  one-line comment noting the asymmetry suffices — no eviction machinery needed.
- **One-line `pflow guide` note:** focus/frame/clear now reach a tab that was hidden when the command
  fired (latched); `select-run` still does not.

## Optional follow-ups (offer, don't block)

- A short **ADR** capturing "close-on-hidden + Point latch" over SharedWorker/HTTP-2/naive-latch (the
  decision is hard-ish to reverse, surprising without context, and a real trade-off).
- Add **"Point latch"** to `context/CONTEXT.md` if it's treated as a lasting domain term.
- Fix the `pflow ui X --run Y` reuse-guard ≤15 s gap in `cli/commands/ui.py` (open a fresh pinned tab for
  the unlatched `--run` verb instead of trusting `windows>0`).

PR #554, opened 2026-07-02
feat: human-in-the-loop approval gates + escalation (Task 125)

spinje:
## Summary

Adds blocking human-in-the-loop gates to pflow workflows: an author-declared
`approval: required` on any step pauses the run for a yes/no before that step
fires, and a `claude-code` agent step can raise a decision **escalation**
mid-run (question + options + recommendation) for a human to answer. Both are
one primitive — a `GateRequest` payload — with two triggers.

## Task

125

## Changes

- **Engine substrate** (`runtime/engine/engine.py`): a pre-exec approval gate
  seam (before the `node.start` trace marker — a denied node never appears in
  the trace) and a post-exec escalation seam (after the node's own completion
  trace, before loop re-entry — so `loop:` + `carry:` can fold the human's
  decision into a re-forked agent).
- **`GateRequest`/`GateResolution`** (`core/gate.py`): the payload IS the seam
  (ADR-0009) — the same JSON-native object feeds the TTY prompt, the `gate`
  trace event, and the non-interactive error diagnostic.
- **Resolver** (`execution/gate_prompt.py`): one builder for every surface —
  CLI interactive (stdin+stderr TTY, not stdout — `pflow wf | jq` still
  gates), MCP (auto-approve only), parallel-batch worker (stub). Ctrl-C at a
  prompt exits 130 like any interrupt, never a node failure.
- **Denial as a clean stop, not a failure**: `WorkflowStatus.DENIED`, CLI exit
  code 3 (vs 1 for failures), its own text/JSON rendering, amber (never
  green) in the web UI badge/banner/run-selector.
- **`--auto-approve=<node-id>`** (repeatable, per-node only — no blanket
  flag, by design) + MCP `workflow_execute(auto_approve=[...])`.
- **Dry-run parity**: gated steps render `[type, approval]` + a footer line
  and `"approval": true` in JSON dry-run — an agent's gate-discovery surface
  before any side effect fires. A cache hit never gates, so cached entries
  are correctly never stamped.
- **`gate` trace events** (pause + resolution, disk-only like `node.start`)
  for audit — `resolved_via: prompt|flag` shows which gates a human answered
  live vs. pre-approved.
- Validation rejects `approval:` on batch-host steps (the preview couldn't
  show resolved `${item}` values); escalation from inside a batch item fails
  loudly rather than silently skipping.
- Guide (`pflow guide approval`) + Mintlify docs
  (`docs/how-it-works/approval-gates.mdx`) covering both gate kinds, the
  agent-operator playbook (discover via dry-run, ask your human, scoped
  `--auto-approve`), and the escalation continue-via-loop recipe.

## Explanation

pflow workflows previously ran start-to-finish with no checkpoint — any
step that posts to Slack, deploys, or otherwise acts on the user's behalf
fired without review. This closes that gap for the **blocking** case (human
present at a terminal during the run); durable, answer-later gates are a
separate follow-up (Task 171) that rides the same `GateRequest` payload
unchanged.

Design highlights carried through review: denial is modeled as a human
verdict (not a node failure) end-to-end — no `error`-successor routing, a
distinct exit code, and an explicit non-success-but-not-failure status
threaded through every consumer (CLI text/JSON, trace trailer, web badge,
run-selector). The escalation trigger is a reserved `result.escalation` key
(works for any node type, not just `claude-code`) rather than a new node
type, keeping the node count and IR surface unchanged. Auto-approve is
deliberately scoped per-node with no blanket form — an agent that always
passes an approve-all flag would defeat the gate's purpose; the friction of
naming each gate is the point.

A 6-agent deep-review pass (silent-failures, impact-completeness,
feature-interactions, agent-ux, test-fidelity, simplicity) found 6 confirmed
issues, all fixed: the denied JSON document now matches the unified failure
shape (`errors`/`diagnostics` arrays, not just a bespoke `gate` key); the
unmatched-`--auto-approve`-id warning no longer misreads a legitimate
nested-workflow gate as a typo; cached-but-gated plan entries are no longer
stamped `approval: true` (a cache hit never gates); the web `RunSelector`
picked up the same amber-denied treatment as the canvas banner and run
callout; the non-interactive pre-flight warning no longer fires for gates
`--only` can't reach; and one v1 limitation (gate events inside a batch
item's sub-workflow don't reach the trace — the item runs on a buffered
collector) is now documented rather than silently dropped.

## Created Docs

- `.taskmaster/tasks/task_125/task-125.md`
- `.taskmaster/tasks/task_125/implementation/implementation-plan.md`
- `.taskmaster/tasks/task_125/implementation/progress-log.md`
- `docs/how-it-works/approval-gates.mdx` (new, plus nav entry in `docs/docs.json`)
- `src/pflow/guide/features/approval.md` (new agent-facing guide topic)

## Testing

- `test_approval_gate.py` (37) — approve/deny/non-interactive, masked
  diagnostics, cache-never-gates, per-iteration loop prompts, escalation
  shape ladder, batch guards, sub-workflow boundary, denied-status contract,
  nested-resolver propagation
- `test_gate_trace.py` (9) — pause/resolution lines, payload round-trip,
  disk-only invariant, denied/failed trailers, child-gate stream
- `test_gate_prompt.py` (15, new) — resolver contract for every surface,
  Decision-14 TTY rule, secret masking/truncation, Ctrl-C to KeyboardInterrupt
- `test_approval_gate_cli.py` (9, new) — exit 3 plus JSON denied document,
  `--auto-approve`, nested-gate-aware warnings, non-interactive fail+warn,
  `-p` suppression, `--only` scoping
- `test_plan_drift.py` (+3) — mutation-verified (both sides) drift pin
  "plan says would-pause equals engine pauses", including the cached-gated
  corner and a gated sub-workflow/loop node
- `test_cli_mcp_parity.py` (+2) — MCP loud-fail with remediation ladder,
  `auto_approve` off-main-thread via `asyncio.to_thread`
- `RunProgress.test.tsx` (+1) — denied renders amber, never the green
  success fallthrough
- Manual: real pty verification of the TTY approve/deny/escalation prompts;
  `pflow ui` screenshot of a replayed denied run (amber banner and badge)
- `make test` 8367 passed, `make test-e2e` 43 passed, `make check` green,
  web `tsc --noEmit` clean, 694 vitest passed

### DIFF
--- a/.taskmaster/tasks/task_125/implementation/implementation-plan.md
+++ b/.taskmaster/tasks/task_125/implementation/implementation-plan.md
@@ -315,15 +315,14 @@ CLI's `finally` finalizes the trailer.
 ### 3.2 Status ripple (complete consumer map — every site verified)
 
 - `WorkflowStatus.DENIED = "denied"` (`core/workflow/status.py:6-21`; str-Enum, additive).
-- Runner **[review-fix — both exceptions]**: dedicated arms in `run()` BEFORE the generic
-  `except Exception` (`runner.py:185-189`): `except GateDenied` →
-  `ExecutionResult(success=False, status=WorkflowStatus.DENIED, ...)` with a denial
-  diagnostic carrying the `GateRequest`; `except GateNotInteractiveError` →
-  `ExecutionResult(success=False, status=FAILED, diagnostics=e.to_diagnostics())` (payload
-  intact — never through the generic `_exception_to_result` flattening). MUST convert to
-  results (not propagate): the CLI's trace finalize at `run.py:340-347` only runs when
-  `result` is non-None — propagating would leave an `incomplete` trace. `_determine_status`
-  (`runner.py:626-642`) never sees these (exception path) — no change there.
+- Runner **[AS-IMPLEMENTED — simpler than reviewed]**: NO dedicated except arms. The generic
+  `_exception_to_result` was verified (session-1 pin) to preserve payload diagnostics via
+  `exception_to_diagnostics` → `to_diagnostics()` — it never flattened them — so DENIED is a
+  one-line status derivation inside it (`isinstance(exception, GateDenied)`), and
+  `GateNotInteractiveError` needed zero runner changes. Both still convert to results (not
+  propagate): the CLI's trace finalize at `run.py:340-347` only runs when `result` is
+  non-None. `_determine_status` (`runner.py:626-642`) never sees these (exception path) —
+  no change there.
 - CLI `_display_execution_result` (`run.py:358-394`): new first branch on
   `result.status is WorkflowStatus.DENIED` → denial output + `ctx.exit(3)`.
   **Denied output is format-aware** **[review-fix]**: text mode → the prose denial line;
@@ -417,9 +416,13 @@ Long preview values: truncate in the renderer only (e.g. 200 chars + `… (N cha
 
 ### 3.4 Flags
 
-- `--auto-approve` (`multiple=True`) in `run.py` → `RunnerConfig.auto_approve:
-  tuple[str, ...] = ()` (`execution/result.py:13-32`, frozen, shared CLI/MCP) → runner builds
-  the resolver. Unknown node-id in the flag: warn at start, don't fail — match against ALL
+- `--auto-approve` (`multiple=True`) in `run.py`. **[AS-IMPLEMENTED]** NO
+  `RunnerConfig.auto_approve` field: the resolver travels as a `runner.run(gate_resolver=...)`
+  kwarg mirroring `progress_callback` end-to-end (CLI and MCP each build their own via
+  `build_gate_resolver`; the CLI pre-flight warnings scan `ir_data` CLI-side in
+  `_prepare_gate_resolver`). Also add the flag to `_PFLOW_FLAGS` (run.py:615 — the
+  misplaced-flag guard pin enforces this). Unknown node-id in the flag: warn at start,
+  don't fail — match against ALL
   top-level node ids (a child gate id is legitimate but invisible here; phrase as
   "no top-level step named X") with a fuzzy suggestion via
   `core/suggestion_utils.find_similar_items` and the list of gated top-level ids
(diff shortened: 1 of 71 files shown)


spinje:
## Review findings addressed (commit 9fe66c96)

Evaluated both reviews with 4 parallel verification agents + direct reads of the two most consequential fix sites. **5 confirmed, 1 disputed, 0 needs-investigation.** All 5 fixes mutation-verified (temporary edit + revert per fix, confirming each test actually catches its bug).

**1. Orphaned trace event on a nested gate stop** (Codex P1) — confirmed, real mechanism was subtly different than described. `WorkflowExecutor.descend()` reserves a seq for a sub-workflow host; the host's own completion event is normally written at engine step 16, but the gate-exception arm re-raises instead of returning, so step 16 never runs — leaving an earlier sibling step's event orphaned. Actual impact: not a later disk-read crash, but `finalize()` itself raising in-memory *before* writing the `run.complete` trailer, silently swallowed by `contextlib.suppress` — the trace file was silently lost. There's also a latent hard-crash surface via `error_formatter.collect_llm_calls()`. Fixed by recording the host's own event (`success=True`, since the run's denied/failed verdict is carried independently via `gate_outcome`) before re-raising — closes the gap at every nesting level for free.

**2. Gate preview masking doesn't recurse into dict/list values** (both reviews flagged this independently, Codex P1) — confirmed. A secret nested under a non-top-level key (`headers: {Authorization: ...}`) rendered unmasked in the TTY prompt and in `GateNotInteractiveError`'s MCP/JSON-visible diagnostic. Fixed by reusing the existing recursive `sanitize_parameters()` sanitizer for non-string preview values.

**3. Auto-approve echo races a parallel-batch worker thread** (Warning) — confirmed. The visibility echo touched the real `OutputController` unconditionally, including from a worker thread (`allow_prompt=False`) — exactly the race the per-worker progress buffer exists to prevent. Now suppressed when `allow_prompt=False`.

**4. Batch-of-sub-workflow dry-run drops the child's `approval` flag** (P2) — confirmed. `_aggregate_batch_child_plans` built fresh synthetic `PlanEntry` objects without forwarding `approval`, so dry-run silently under-reported a gate the engine still fires on at runtime. Fixed.

**5. `sorted()` on a set that could contain `None`** (Suggestion) — confirmed as a latent property; verified unreachable today (the parser always sets `"id"`), but hardened for free.

**Disputed**: the docs example placing `--auto-approve` after the workflow name (P2) — verified empirically correct as written (`allow_interspersed_args=True` + it being a recognized Click option means position doesn't matter). No change.

Gates: `make test` 8376 passed (+9), `make test-e2e` 43 passed, `make check` fully green.

Issue #413, opened 2026-05-18
LiteLLM mutates pflow's system_blocks list in-place — trace and downstream consumers see corrupted state


spinje:
## Background

During live verification of the multi-breakpoint prompt caching feature (PR #412), I discovered that pflow's trace records `llm_system` content that doesn't match what pflow actually placed. For a 4-chunk Anthropic workflow this is invisible (Anthropic preserves the markers, so the trace happens to match). For OpenAI / Gemini workflows it's significantly misleading: the trace shows zero `cache_control` markers even though pflow placed one.

The root cause turns out not to be a trace issue — it's that **LiteLLM mutates pflow's data in place** when normalizing the request for non-Anthropic providers.

## Reproduction

Direct test isolating the mutation (`/tmp/check_post_complete.py` during verification):

```python
system_blocks = [
    {"type": "text", "text": "Block A..."},
    {"type": "text", "text": "Block B..."},
    {"type": "text", "text": "Block C..."},
    {"type": "text", "text": "Block D...", "cache_control": {"type": "ephemeral"}},
]

print("BEFORE complete():", [("cache_control" in b) for b in system_blocks])
# → [False, False, False, True]

try:
    complete(model="openai/gpt-4o-mini", system=system_blocks, prompt="...")
except MissingApiKeyError:
    pass  # the error fires AFTER the mutation has already happened

print("AFTER complete():", [("cache_control" in b) for b in system_blocks])
# → [False, False, False, False]   ← block 3 lost its cache_control key
```

The mutation is fully consistent: every call to `complete()` for a non-Anthropic model strips `cache_control` keys from pflow's list before sending. The marker is gone from pflow's own data structure after the call.

## What pflow does (the data path)

1. `_build_system_blocks` (`src/pflow/nodes/llm/llm.py:769-862`) constructs the list of content blocks with `cache_control` markers placed by `compute_marker_chunk_indices`.
2. `_strip_below_min_cache_markers` walks the list to remove markers below the provider minimum.
3. The adapter `complete()` (`src/pflow/core/llm_client.py`) passes this list directly to `litellm.completion(system=system_blocks, ...)`.
4. LiteLLM, internally, normalizes the request for the target provider. For OpenAI/Gemini, that includes `del block["cache_control"]` on the input list.
5. pflow's trace hook (`src/pflow/runtime/workflow_trace.py:606`) registered earlier stored a REFERENCE to that same list (`self.llm_systems[node_id] = system`).
[...]

PR #588, opened 2026-07-13
fix(mcp): strip _*_source_line sidecar from forwarded tool args (#497)

### LINKED ISSUE #497: Fenced code blocks leak a `_<param>_source_line` sidecar into forwarded params (breaks MCP nodes)
## Summary

Any node param can be supplied via a fenced code block (the parser already routes ` ```lang <param> ` blocks to `params[<param>]`). But the parser/compiler inject a **`_<param>_source_line` sidecar** key into the node's params for error-line mapping, and nodes that **forward their params** — notably the **MCP node** — pass that sidecar straight through to the underlying call, which rejects it.

Net effect: you can't use a fenced code block for a param on an MCP node (and any future param-forwarding node), even though the parser supports it.

## Reproduction

An MCP node with a param supplied via a fenced block:

````markdown
### settle
- type: mcp-chrome-devtools-evaluate_script

```javascript function
async () => { return "hi"; }
```
````

**Actual:**

```
Unknown argument for tool "evaluate_script": "_function_source_line".
Expected arguments: "function", "args", "filePath", "dialogAction".
```

**Expected:** the block supplies `function`; the tool call succeeds.

Workaround today: a YAML block scalar `- function: |`, which doesn't trigger the sidecar — but fenced blocks should work too, for parity with the `code` node and for syntax highlighting.

## Root cause (verified, with `file:line`)

1. **Parser** records the block's line in a per-node `_source_lines` dict — `src/pflow/core/markdown_parser.py:1201-1202`:
   ```python
   params[block.param_name] = block.content
   source_lines = node.setdefault("_source_lines", {})
   source_lines[block.param_name] = block.start_line + 1
   ```
2. **Compiler** flattens that into per-param keys `_<param>_source_line` and merges them into `params` — `src/pflow/runtime/compilation/compiler.py:285-287`:
   ```python
   source_lines = node_data.get("_source_lines")
   if source_lines:
       params = {**params, **{f"_{k}_source_line": v for k, v in source_lines.items()}}
   ```
   These survive `split_params` into `static_params` (kept deliberately — `template_resolution.py:96-98`) and land on `node.params`.
3. **MCP node** forwards every param to the tool, filtering only **double**-underscore keys and `timeout` — `src/pflow/nodes/mcp/node.py:153`:
   ```python
   tool_args = {k: v for k, v in self.params.items() if not k.startswith("__") and k != "timeout"}
   ```
   `_function_source_line` starts with a **single** `_`, so it isn't filtered → forwarded verbatim to the tool call (`node.py:205-212` / `:285` / `:347`) → rejected.

**Why the `code`/`python` node is immune:** it reads params *by name* (only `code`, `timeout`, `inputs`, `requires`) and never iterates `self.params` to forward arbitrary keys — `python_code.py:856`. It even consumes the sidecar on purpose for error-line mapping (`python_code.py:660`, used at `:561-563`). So the sidecar is a *feature* for name-reading nodes and a *leak* for param-forwarding nodes.

## Proposed fix

The sidecar is pflow-internal metadata and should never reach a node's outbound payload. There's already a correct, suffix-based predicate to mirror — `src/pflow/runtime/engine/instrumentation.py:166` (`not k.endswith("_source_line")`).

- **Minimal (resolves the observed bug):** add `and not k.endswith("_source_line")` to the `tool_args` comprehension at `nodes/mcp/node.py:153`. Keeps `MCPNode` server-agnostic — it strips *pflow's own* metadata, not anything server-specific (consistent with `nodes/CLAUDE.md:146-153`).
- **General (matches "any param on any node"):** promote the sidecar-suffix knowledge into one shared helper/constant (today `_METADATA_KEY_SUFFIXES` is module-private in `cache.py:30-34`) — e.g. `strip_internal_params(params)` — and use it wherever a node forwards a params dict, so the whole class is covered, not just MCP. **Recommended**, since the goal is universal fenced-block param support.

## Acceptance

- An MCP node with a fenced-block param runs (repro passes); no `_*_source_line` reaches the tool.
- `code`/`llm` error-line mapping still works (sidecar still readable by name).
- Cache keys unaffected (`instrumentation.py:166` already strips the suffix).
- Regression test: an MCP node with a `_x_source_line` param asserts it's excluded from `tool_args`.

## Note

The accurate key form is `_<param>_source_line` (single leading underscore, e.g. `_function_source_line`) — there is no `function_source_line` variant in the code.

spinje:
## Summary

Fenced code blocks leaked a pflow-internal `_<param>_source_line` sidecar into the MCP node's forwarded tool arguments, so any MCP node param supplied via a ```` ```lang <param> ```` block was rejected by the server (`Unknown argument "_function_source_line"`).

Fixes #497

## Changes

- `src/pflow/nodes/mcp/node.py` — the `tool_args` comprehension now also filters `not k.endswith("_source_line")`, so the sidecar never reaches the tool call.
- `tests/test_nodes/test_mcp/test_source_line_sidecar.py` — new regression tests (all fail without the fix).

## Explanation

Fenced-block params get a `_<param>_source_line` sidecar injected for error-line mapping: the parser records the line in `node["_source_lines"]` and the compiler flattens it into `_<param>_source_line` params (`compiler.py:285-288`). Nodes that read params **by name** are immune and some depend on the sidecar on purpose (`python_code.py` reads `_code_source_line`; `template_resolution.py` reads `_<key>_source_line` for error rendering). The MCP node is the **only** node that forwards its params **wholesale** (iterating `self.params`), so it was the only leak — confirmed by an impact-completeness sweep of every node and every param-forwarding path (sub-workflow inputs, batch, graph build, display).

The fix is therefore targeted (strip at the MCP forward filter), not a global strip that would break the by-name readers. No shared helper: MCP is the single wholesale-forwarding site, and the three sibling `_source_line` predicates (`instrumentation.py`, `cache.py`, here) operate on different data shapes with different filter sets — consolidating would move complexity, not concentrate it (deletion test).

## Testing

- `TestSourceLineSidecarFiltering::test_source_line_sidecar_excluded_from_tool_args` — a `_function_source_line` param is dropped from `prep()` `arguments`; the real `function` param survives.
- `TestSourceLineSidecarFiltering::test_only_source_line_suffix_is_stripped` — a user param merely starting with `_` (`_leading_underscore`) is preserved; only the `_source_line` suffix is internal (guards against a lazier `startswith("_")` fix).
- `TestFencedBlockParamEndToEnd::test_fenced_block_param_does_not_leak_sidecar_to_tool_args` — the issue's repro driven through the **real** parser → compiler → `prep()` stack; asserts the compiler injects `_function_source_line` (precondition, so the test can't go vacuous) and that it's stripped from the outbound `arguments`.
- All three fail on the unpatched node and pass with the fix. `make check`, `make test` (8797 passed), and `make test-all-local` (8840 passed, 1 skipped) green.

## Note on extra commits

This branch was cut from a local `main` tip (`19bfbbba`) that isn't on `origin/main` yet, so its orchestration-doc changes (`CURRENT-STATE.md`, `session-04.md`) appear in this diff and ride to main with the merge — expected for the worktree flow.

One extra `chore` commit (`[skip review]`) regenerates a **pre-existing** stale Codex mirror asset (`.agents/skills/start-orchestration/SKILL.md`) that made `make check`'s asset-sync verification red on every branch based on this main (the command source was edited on main without regenerating its mirror). It is unrelated to #497 and safe to drop if fixed on main separately.

### DIFF
--- a/.agents/skills/start-orchestration/SKILL.md
+++ b/.agents/skills/start-orchestration/SKILL.md
@@ -30,7 +30,7 @@ between what documents claim and what code does.
    merge, or reconcile without it), then `CURRENT-STATE.md` + the LATEST `sessions/session-NN.md`
    — if that file is thin (a short check-in, an aborted session), read one further back until you
    hit a substantive one (DECISIONS #10) — then `BRAINDUMP.md` (the role's tacit layer, refreshed
-   at each close). Treat every claim as a **pointer to verify, not a fact**. Read all these files in full before you continue.
+   at each close). Treat every claim as a **pointer to verify, not a fact**.
 2. Verify reality: `git fetch` + `git log --oneline -15 origin/main` · `gh pr list --state merged
    --limit 10` · `gh issue list --state open --limit 40` (scan for new since the stamp) ·
    `./scripts/tasks` · `git worktree list` · TaskList for live subagents.
--- a/.taskmaster/orchestration/CURRENT-STATE.md
+++ b/.taskmaster/orchestration/CURRENT-STATE.md
@@ -20,15 +20,25 @@ Every claim here is a pointer to verify, not a fact._
   **Opus**, not Fable — scope is planners only; #8 (UI phases → Fable) and #9 (lane-B Fable
   opt-in) stand. DECISIONS #3 amended; ORCHESTRATION + agent def reconciled.
 
-## In flight
+## In flight (launched 2026-07-13, session-04 — 3 parallel lane-B Opus builds)
 
-- ~~**#565 (MCP probe output paths)**~~ — MERGED via PR #583 as `bd6f520a`; all merged-result gates
-  green after #581, worktree/branch safely pruned.
-- ~~**#581 (flaky Windows caplog tests)**~~ — MERGED via PR #584 as `08b319ca`; issue-specific Windows
-  gate passed, authorized unrelated-flake rerun passed, worktree/branch safely pruned.
-- ~~**#585 (Windows MCP config replace WinError 5)**~~ — MERGED via PR #586 as `2ee5fefc`; all
-  Windows gates green, worktree/branch safely pruned. Verified cause: Windows replacement can
-  transiently reject despite existing process-wide `RLock`; fix narrowly retries WinError 5/32.
+Surfaces verified against main by 3 searchers before launch; collision matrix confirmed disjoint
+(files + semantics); none touch `runtime/engine/`/`workflow_executor` or trace format.
+
+- **#413** LiteLLM mutates pflow's `system_blocks` in place → trace lies for OpenAI/Gemini cache
+  runs. Fix = deep-copy at the one adapter boundary (`core/llm_client.py:358`, before
+  `litellm.completion`); Option A (issue's Option C sidecar-refactor rejected as over-eng).
+- **#497** `_*_source_line` sidecar leaks through MCP node's forward filter → `Unknown argument`.
+  Fix = one line at `nodes/mcp/node.py:153` (`and not k.endswith("_source_line")`, matching the
+  `instrumentation.py:166` precedent). No shared helper (one forwarding node → deletion test).
+- **#183** secrets embedded in shell `command`/`stdout`/`stderr` leak to CLI **and** MCP error
+  output. Single shared seam = `_enrich_error_from_node_output` (`executor_service.py:354-359`);
+  covers both. TRAP: `sanitize_parameters` redacts by KEY — a secret inside a string value needs
+  **value-level** substring redaction of known-secret values (env/sensitive param values). Scope
+  = shell fields only (HTTP already redacted; consolidation is a separate follow-up).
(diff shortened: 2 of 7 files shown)


spinje:
Addressed the Codex P2 ("Preserve real MCP args ending in `_source_line`") in commit `2f167c5f`.

The filter was too broad (`endswith("_source_line")` would also drop a genuine tool argument such as `data_source_line`). It is now precise: the compiler emits the sidecar as `f"_{base}_source_line"` where `base` is a real param, so a key is stripped only when its inner base name is itself present in `self.params`. Real args like `data_source_line` (no leading `_`) or `_x_source_line` (no `x` param) are forwarded intact. See `_is_source_line_sidecar` in `src/pflow/nodes/mcp/node.py`.

New regression test `test_real_arg_ending_in_source_line_is_preserved` proves both directions and is mutation-verified to fail under the old broad filter.

Also merged current `main` (#413) into the branch; `make check` + `make test` (8800 passed) green on the merged result.

PR #596, opened 2026-07-15
fix(cli): describe local workflow files

### LINKED ISSUE #590: describe: support local workflow files, not just saved workflows
## Problem

`pflow describe` only resolves **saved** workflows (by library name). Pointing it at a workflow **file path** fails, even though the file is a valid, parseable workflow:

```
$ pflow describe examples/real-workflows/git-worktree-task-creator/workflow.pflow.md
Error: Workflow 'examples/...workflow.pflow.md' not found.
```

This is an unnecessary asymmetry: `pflow <path>` runs a file, `pflow --validate-only <path>` validates a file, `pflow guide <path>` auto-detects topics from a file — but `pflow describe <path>` can't show its interface. Many workflows (examples, in-repo `.pflow.md` files, drafts under review) are never saved to the library, so `describe` is unavailable exactly where an agent/human most wants a quick "what are the inputs/outputs and how do I call this" view.

## Current behavior (grounding)

- `src/pflow/cli/commands/describe.py` — takes `NAME`, calls `WorkflowManager.exists(name)` / `.load(name)`, then `format_workflow_interface(name, metadata)`. No file-path branch; a path just fails `exists()`.
- `src/pflow/execution/formatters/workflow_describe_formatter.py:format_workflow_interface(name, metadata)` — reads only `metadata["ir"]` and `metadata["description"]`, plus **optional** execution-history fields (`_format_execution_history_section` returns `""` when absent). So the formatter already tolerates missing history.

## Why this is a small change

A unified resolver already exists and already handles file paths:

- `src/pflow/execution/workflow_resolver.py:resolve_workflow(identifier, wm=None) -> ResolvedWorkflow` — for a file path returns `ResolvedWorkflow(ir=..., source="file", file_path=<abs>, title=..., description=...)`.

So `describe` can resolve **either** a saved name **or** a file path, then adapt the result into the metadata shape the formatter already expects. Roughly:

1. If `WorkflowManager.exists(name)` → keep today's path (saved workflow, includes execution history).
2. Else → `resolve_workflow(name)`; on success build `metadata = {"ir": resolved.ir, "description": resolved.description or "No description"}` (no history fields) and format that.
3. Preserve the current not-found error + "Did you mean" suggestions only when the argument is neither a saved name nor a resolvable file.

## Behavior notes / edge cases

- **Execution history**: unsaved files have none — the history section is already conditional, so it simply won't render. (Consider a one-line hint like "not saved — no run history; `pflow save` to track it".)
- **Example-usage line**: generated from input names, so it works for files too (the example would show the file path or a placeholder name — decide which).
- **Inline/content workflows**: out of scope; `describe` takes a positional arg, keep it to saved-name + file-path.
- **CLI routing**: `describe` is a registered subcommand, so `pflow describe <path>` already routes here — only the resolution logic needs to change.

## Acceptance criteria

- [ ] `pflow describe <path/to/workflow.pflow.md>` prints the interface (name, description, inputs, outputs, example usage) for an unsaved file.
- [ ] `pflow describe <saved-name>` is unchanged, including the execution-history section.
- [ ] A truly unknown argument (neither saved name nor existing file) still errors with the "Did you mean" suggestions.
- [ ] Tests cover: saved-name (regression), file-path success, and unknown-arg error. See `tests/test_cli/test_workflow_commands.py` (patches `pflow.cli.commands.describe.WorkflowManager`).

## Out of scope

- Making `describe` summarize/shorten prose — it is a verbatim pass-through of the workflow's own input/output descriptions by design; density is a source-side concern, not a `describe` concern.

spinje:
Closes #590

Summary:
- resolve local workflow files through the existing unified resolver while preserving saved-workflow history and unknown-name suggestions
- preserve the typed path in the heading and emit shell-copyable usage, including paths with spaces
- make shared CLI path routing and adjacent guidance consistent, with reference docs and regressions

Verification:
- make check
- make test-all-local: 9059 passed, 2 skipped
- mandatory manual-verification skill: copied spaced-path example executed successfully
- final deep-review re-reviews: no Critical or Warning findings

### DIFF
--- a/docs/reference/cli/describe.mdx
+++ b/docs/reference/cli/describe.mdx
@@ -1,13 +1,19 @@
 ---
 title: "pflow describe"
-description: "Show a saved workflow's interface"
+description: "Show a saved workflow or local workflow file's interface"
 icon: "file-search"
 ---
 
 ## Usage
 
 ```bash
-pflow describe <WORKFLOW>
+pflow describe <NAME_OR_PATH>
 ```
 
-Shows inputs, outputs, history summary, and example usage for a saved workflow.
+Shows inputs, outputs, and a copyable example command. Saved workflows also include their execution history; unsaved files omit history.
+
+```bash
+pflow describe my-saved-workflow
+pflow describe ./drafts/release-notes.pflow.md
+pflow describe './draft workflows/release notes.pflow.md'
+```
--- a/docs/reference/cli/index.mdx
+++ b/docs/reference/cli/index.mdx
@@ -23,7 +23,7 @@ pflow [command] [options] [arguments]
     Run workflows by name or file
   </Card>
   <Card title="pflow list / find / describe" icon="workflow" href="/reference/cli/list">
-    Find and inspect saved workflows
+    Find saved workflows and inspect them by saved name or file path
   </Card>
   <Card title="pflow skill" icon="wand-magic-sparkles" href="/reference/cli/skill">
     Publish workflows as AI agent skills
--- a/src/pflow/cli/commands/describe.py
+++ b/src/pflow/cli/commands/describe.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import shlex
 import sys
 
 import click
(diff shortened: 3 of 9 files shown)


spinje:
Automated review disposition:

1. Addressed in `669110fb`: both CLI callers now use one shared command-seam adapter. It preserves the formatter interface, renders the raw heading, uses the shell-safe example name, and fails loudly if the formatter heading contract changes.
2. Addressed in `669110fb`: `_handle_workflow_not_found` is now annotated `NoReturn`.
3. Intentionally unchanged: a `.pflow.md` argument denotes a file, so a missing file does not fall back to a stripped saved-workflow name and silently lose saved history. This behavior is regression-tested.

Follow-up verification: `make check`; `make test-all-local` (9,059 passed, 2 skipped).

PR #599, opened 2026-07-17
fix(llm): unwrap nested structured responses

### LINKED ISSUE #598: output_schema: parsed response can arrive wrapped as {"json_tool_call": {...}} on Anthropic models — schema not enforced post-parse
## Summary

With `output_schema` set on an `llm` node, the parsed `response` can intermittently arrive as
`{"json_tool_call": {...the actual schema-valid object...}}` — a single-key wrapper dict whose
key is **LiteLLM's internal response-format tool name**, not part of the user's schema. The
wrapper violates the schema's own top-level `required` list, yet flows into downstream nodes
unparsed and unvalidated. This breaks the documented contract ("guarantees valid JSON via
constrained decoding", llm guide) and every `${node.response.<key>}` template reference
downstream of the call.

Observed once in ~4 live runs of the same workflow on `anthropic/claude-sonnet-5`; the runs
minutes before and after parsed clean — intermittent, model-output-shape dependent.

## Environment

- pflow 0.14.0 (`pflow-cli`, local checkout)
- litellm 1.86.1
- model: `anthropic/claude-sonnet-5`
- macOS (darwin 24.6.0)

## Steps to reproduce

Intermittent by nature (depends on the model's tool-call emission shape), but the setup that
produced it:

1. An `llm` node with:
   - a large-ish `output_schema` (object, ~20 properties, 6 required top-level fields, nested
     arrays of objects),
   - `max_tokens: 16000`,
   - a long system prompt (~29k input tokens) + a prompt file with `${...}` substitutions.
2. Run the workflow repeatedly with `--no-cache` (the failing run was NOT a cache replay —
   fresh `output_tokens: 1441`, `uncached_input_tokens: 28811`).
3. On the failing run, the node's `response` was a dict with **exactly one key**:

   ```json
   {"json_tool_call": { "...": "the complete, otherwise schema-conforming object" }}
   ```

   The inner object was complete and valid against the schema; only the wrapper was wrong.

A downstream `code` node reading `response` keys then failed (in our case a guard raised on
the missing `questions` key; without the guard, every `${synthesize.response.<field>}`
template reference in the workflow file would have failed instead).

Debug trace of the failing run (local): `~/.pflow/debug/workflow-trace-f9757ce5-precall-brief-20260717-001348-384332.json`
(synthesize node event → `node_output.response` shows the single-key wrapper; `llm_call`
shows the fresh-token accounting above). Happy to attach relevant excerpts.

## Where the wrapper comes from (root-cause analysis)

- pflow passes `schema` to LiteLLM as
  `response_format={"type": "json_schema", "json_schema": {"name": "response", "schema": ..., "strict": True}}`
  (`src/pflow/core/llm_client.py:302-305`).
- For Anthropic, LiteLLM implements `response_format` by injecting a **forced tool call**
  named `json_tool_call`:
  - `litellm/constants.py:1286` — `RESPONSE_FORMAT_TOOL_NAME = "json_tool_call"` ("default
    tool name used when converting response format to tool call");
  - `litellm/llms/anthropic/chat/transformation.py:1491,1590` — tool creation;
  - `.../transformation.py:1999` `_resolve_json_mode_non_streaming` — strips the internal
    tool call and returns its `arguments` as the message content.
- Anthropic **does not grammar-constrain tool-call arguments server-side** — the model's
  `input` for the forced tool is best-effort against the tool's `input_schema`. The model can
  therefore occasionally double-nest its answer under the visible tool name *inside the
  arguments* (i.e. `arguments = {"json_tool_call": {...}}`), which is exactly the shape we
  received: LiteLLM's extraction strips the tool-call layer correctly, but the arguments
  themselves carry a second wrapper.
- Neither LiteLLM's extraction nor pflow's llm node **re-validates the parsed dict against
  the schema** after parsing (`llm.py` parses JSON and returns the dict; the "raw text on
  parse failure" fallback only covers JSON *syntax* failure, not schema violation).

So "constrained decoding" on the Anthropic path is actually "forced tool call with
best-effort arguments", and there is no post-parse schema check to catch the gap.

## Expected

`response` is either (a) a dict that validates against `output_schema`, or (b) the documented
parse-failure fallback (raw text + `error` set) so workflows fail loudly at the node that
broke, not at an arbitrary downstream template reference.

## Suggested fixes (either closes it; both would be defense-in-depth)

1. **Post-parse schema validation in the llm node**: after JSON parsing, validate against
   `output_schema` (jsonschema is already transitively available via litellm). On violation:
   retry once (the same repair semantics as a parse failure), and if still invalid, populate
   `error` + fall back to raw text — matching the documented contract.
2. **Targeted unwrap shim**: if the parsed dict is exactly
   `{"json_tool_call": <dict>}` (single key == litellm's `RESPONSE_FORMAT_TOOL_NAME`) and the
   inner dict validates where the outer doesn't, unwrap it (log a warning). Cheap, catches
   the known model habit; option 1 is the principled fix.

Also worth a docs touch: the llm guide's "guarantees valid JSON via constrained decoding"
overstates the Anthropic path until one of the above lands.

## Workaround used meanwhile

A code node immediately after the llm node unwraps a single-key `json_tool_call` dict (with a
run-summary warning) before anything reads `response` keys, and all downstream reads go
through that node's normalized result instead of `${<llm-node>.response.*}`.

Sanjays2402:
Closes #598.

LiteLLM can return otherwise valid structured output nested under its internal `json_tool_call` name. Unwrap that shape only when the outer value fails the requested schema and the inner value passes it; all other responses remain unchanged.

Test: `pytest -q tests/test_nodes/test_llm/test_llm.py` (81 passed)

### DIFF
--- a/src/pflow/nodes/llm/llm.py
+++ b/src/pflow/nodes/llm/llm.py
@@ -9,6 +9,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
+from jsonschema.validators import validator_for
+
 sys.path.insert(0, str(Path(__file__).parent.parent.parent.parent.parent))
 
 from pflow.core.cache_ttl import is_cache_ttl_supported_by_provider, parse_cache_ttl
@@ -1342,7 +1344,20 @@ class LLMNode(Node):
         # error_class consistently with the _call_llm path.
         if exec_res["has_schema"]:
             try:
-                shared["response"] = json.loads(raw_response)
+                parsed_response = json.loads(raw_response)
+                output_schema = prep_res.get("output_schema")
+                if (
+                    isinstance(parsed_response, dict)
+                    and set(parsed_response) == {"json_tool_call"}
+                    and isinstance(parsed_response["json_tool_call"], dict)
+                    and isinstance(output_schema, dict)
+                ):
+                    validator = validator_for(output_schema)(output_schema)
+                    inner_response = parsed_response["json_tool_call"]
+                    if not validator.is_valid(parsed_response) and validator.is_valid(inner_response):
+                        logger.warning("Unwrapped structured output nested under LiteLLM's internal tool name")
+                        parsed_response = inner_response
+                shared["response"] = parsed_response
             except json.JSONDecodeError as e:
                 # Build the same error dict shape as _call_llm so the runtime
                 # path produces the same structured Diagnostic. Use
--- a/tests/test_nodes/test_llm/test_llm.py
+++ b/tests/test_nodes/test_llm/test_llm.py
@@ -915,6 +915,32 @@ class TestStructuredOutput:
         assert isinstance(shared["response"], dict)
         assert shared["response"] == {"name": "Alice", "age": 30}
 
+    def test_structured_response_unwraps_litellm_tool_name(self, monkeypatch):
+        """A schema-valid value nested under LiteLLM's internal tool name is recovered."""
+
+        def custom_complete(**kwargs):
+            return AdapterResponse(
+                text='{"json_tool_call":{"name":"Alice","age":30}}',
+                usage={},
+                model=kwargs.get("model", "test"),
+                has_schema=True,
+            )
+
+        monkeypatch.setattr("pflow.nodes.llm.llm.complete", custom_complete)
+
+        node = LLMNode()
+        node.set_params({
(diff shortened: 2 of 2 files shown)


spinje:
Thank you for your contribution but this issue was already fixed!

Issue #606, opened 2026-08-02
Consolidate the dual LLM provider tables (llm_providers.PROVIDERS vs settings._LLM_PROVIDERS)


spinje:
Two independent provider→env-var tables exist and nothing enforces their agreement:

- `src/pflow/core/llm_providers.py` → `PROVIDERS` — canonical registry (anthropic, openai, gemini). Drives `resolve_provider_api_key` (which now feeds the actual `api_key` sent to LiteLLM, see #605), validator key checks, `MissingApiKeyError` remediation text, and TTS.
- `src/pflow/cli/commands/settings.py` (~line 481) → `_LLM_PROVIDERS` — larger curated display table (25+ providers with `single`/`or`/`and`/`local` semantics) behind `pflow settings llm providers`, with its own `_provider_status` presence logic.

The gemini row is duplicated in both with the same two env vars in the same order — by hand. If they drift, `settings llm providers` (the env-var discovery surface agents read and act on) teaches a configuration that key resolution won't use. Surface-says-X-runtime-does-Y is exactly the failure class behind #605.

**Fix:** derive the three first-class rows in `_LLM_PROVIDERS` from `PROVIDERS` (semantics from `len(env_vars)`: 1 → `single`, >1 → `or`); the curated display-only rows stay as data. Fallback if derivation fights the display schema: a parity test asserting every `PROVIDERS` entry appears in `_LLM_PROVIDERS` with identical env vars and order.

Constraints:
- Env-var **order is semantic** (canonical-first governs which key value is sent) — must be preserved exactly.
- Do **not** promote display-only providers into `PROVIDERS` — entries there carry behavioral weight (explicit api_key pass, validation gates); that's a per-provider feature decision.
- `pflow settings llm providers` output (text + `--output-format json`) should be unchanged.

Existing coverage to keep green: `tests/test_cli/` settings tests, `tests/test_core/test_llm_providers.py`, `tests/test_core/test_llm_config.py::TestResolveProviderApiKey`.

PR #607, opened 2026-08-02
fix: pass the validated API key to LiteLLM and surface provider error text (#605)

### LINKED ISSUE #605: Runtime can silently use GOOGLE_API_KEY instead of the validated GEMINI_API_KEY; provider 404 rendered as 'Unknown model' with no provider text
## Symptom

A workflow with `model: gemini/gemini-2.5-flash` passes `pflow check` (5/5 green) but the first run fails at the first LLM node:

```
Unknown model: gemini/gemini-2.5-flash
Your configured API key supports 'anthropic/claude-sonnet-4-5'.
```

Both keys valid (Gemini key returns 50 models via direct REST; Anthropic key completes). Reported by a real user on 0.15.0.

## Root causes (three independent defects)

1. **Key shadowing** — pflow validates key presence canonical-first (`GEMINI_API_KEY`, then `GOOGLE_API_KEY`) but never passes the key to LiteLLM; LiteLLM re-resolves from `os.environ` with the *reverse* precedence (`litellm/llms/gemini/common_utils.py`). With both vars exported, the call silently uses `GOOGLE_API_KEY` — if that's a generic GCP key without Generative Language API access, Google returns 404. Validate-green, runtime-dead, and no surface names the actual key used. pflow's own surfaces (`settings llm providers` "GEMINI_API_KEY or GOOGLE_API_KEY", the `MissingApiKeyError` "This provider also accepts:" hint) actively encourage setting both.

2. **404 mislabeled** — `_classify_litellm_error` maps every `openai.NotFoundError` to `UnknownModelError("Unknown model: ...")`, but a provider 404 can equally mean key/project entitlement. The model was in the catalogue the whole time.

3. **Provider text never rendered** — `provider_message` (the raw upstream diagnosis, e.g. `models/gemini-2.5-flash is not found for API version v1beta`) is captured on the exception and claimed by a comment in `exceptions.py` to be "rendered above suggestions", but `diagnostic_render.py` never reads it. The one line that would have identified the failure never printed. The static hint "Your configured API key supports 'X'" is just the top of the auto-detect priority ladder (Anthropic always wins when its key exists) and misdirects debugging whenever multiple keys are set.

## Fix

- `complete()` resolves the key via a new `llm_config.resolve_provider_api_key` (canonical-first, env then settings — the same algorithm validation uses, now shared) and passes `api_key=` explicitly for registered providers; unregistered prefixes keep full delegation.
- Render `provider_message` as a `Provider response:` block; mask key material (`?key=`, `AIza…`, `sk-…`) at capture since the text reaches terminal, traces, and JSON.
- Reword the 404 suggestions to keep the entitlement reading open, and replace the misleading hint with an honest "Known-working fallback" line.

Follow-up filed separately: Task 178 (consolidate the dual provider tables in `llm_providers.py` / `settings.py`).

spinje:
## Summary
A validate-green workflow using `gemini/gemini-2.5-flash` died at runtime with `Unknown model` while both provider keys were valid (real 0.15.0 user report). Three independent defects lined up: LiteLLM silently used a different key than the one pflow validated, a provider 404 was mislabeled as a model-name problem, and the provider's own error text — the one line that identified the failure — was captured but never rendered.

Fixes #605

## Changes
- `llm_config.py`: new `resolve_provider_api_key(provider)` — walks env vars canonical-first (`GEMINI_API_KEY` before `GOOGLE_API_KEY`), `os.environ` then settings. `_has_provider_key` is now a one-line derivation of it, so the presence check and the key a call uses can never disagree.
- `llm_client.py`: `complete()` passes the resolved key as `api_key=` for registered providers (caller-supplied `api_key` in `model_options` wins; unregistered prefixes like `openrouter/...` keep full delegation). Also masks key material (`?key=`, `AIza…`, `sk-…`) in provider error text at the classification seam.
- `diagnostic_render.py`: new `Provider response:` block renders `provider_message` (multi-line indented) for any LLM error carrying it.
- `exceptions.py`: 404 suggestions now keep the key/entitlement reading open; the misleading `Your configured API key supports 'X'` hint (just the top of the auto-detect ladder — always Anthropic when that key exists) replaced with an honest known-working-fallback line.
- `llm_providers.py` comment + `core/CLAUDE.md` synced to the new key-resolution reality.

## Explanation
pflow validated key *presence* as a set-union over env-var aliases but delegated key *selection* to LiteLLM, whose Gemini lookup checks `GOOGLE_API_KEY` before `GEMINI_API_KEY` — the reverse of pflow's canonical order. With both vars exported (pflow's own `settings llm providers` output and `MissingApiKeyError` hint encourage exactly that), the request went out with the un-tested `GOOGLE_API_KEY`, Google returned 404, and pflow reported "Unknown model" plus a hint naming Anthropic — misdirecting the user into a key-detection goose chase. Passing the validated key explicitly makes pflow's canonical-first resolution the single authority; the rendering and wording fixes make the next provider-side 404 a one-line read instead.

Masking exists because `provider_message` now reaches the terminal and already reached traces/JSON — Google's REST surface can echo the API key in request URLs.

9 files changed, 337 insertions(+), 22 deletions(-).

Follow-up filed separately: #606 (consolidate the dual provider tables).

## Testing
- `test_canonical_key_wins_over_litellm_precedence` — the incident regression: both Gemini vars set with different values → the `GEMINI_API_KEY` value is sent
- `test_alias_var_used_as_fallback` / `test_no_key_configured_omits_kwarg` — `GOOGLE_API_KEY`-only setups still work; no key → kwarg omitted so LiteLLM's auth-error path is unchanged
- `test_settings_stored_key_resolves_without_env` — non-CLI callers don't depend on `inject_settings_env_vars()`
- `test_caller_supplied_api_key_wins` / `test_unregistered_provider_keeps_delegation` — override and delegation boundaries
- `TestResolveProviderApiKey` (6 tests) — canonical order, settings fallback + whitespace strip, env-over-settings, unknown provider
- `test_unknown_model_error_renders_provider_response_and_honest_suggestions` — end-to-end exception → render: provider text present, old misleading phrase absent
- `test_provider_message_is_masked_through_classifier` + unit test — key material scrubbed
- Full gates on this tree: `make check` clean (ruff, mypy, deptry, hooks); `make test-all-local` (incl. e2e): 9163 passed, 2 skipped

https://claude.ai/code/session_016ToQhpLMjSofEwaQcRdwVG


spinje:
All ten findings from both reviews are addressed in 351649a2.

**Claude review:**
1. *Provider response unbounded* — capped at 5 lines × 300 chars with a `... (truncated — full text in the trace)` tail, matching the file's conventions. Plus a guard for whitespace-only `provider_message` (pre-existing crash shape the review's snippet inherited).
2. *Exact-secret masking* — `complete()` threads `kwargs["api_key"]` into `_classify_litellm_error`; the exact wire secret is replaced verbatim before the shape patterns run. The `LLMResponseParseError` construction site now masks through the same helper, so the CLAUDE.md invariant ("every typed exception carries sanitized text") is true again.
3. *Pattern pairs* — `lastindex` lambda replaced with explicit `(pattern, replacement)` tuples.
4. *Long suggestions* — one action per bullet; the fallback hint is now "To isolate model-vs-key, retry with 'X' (a model your configured keys already reach)" — no auto-detect-ladder internals.
5. *tts duplicate resolution* — `_gemini_api_key()` delegates to `resolve_provider_api_key("gemini")`.
6. *Auth-path mask coverage* — added `test_auth_error_masks_exact_wire_secret` (AuthenticationError → MissingApiKeyError path, odd-shaped secret).

**Codex review:**
- *P1 `api_key=` label* — pattern broadened to `api_key=`/`api-key=`/`key=`; the exact-secret replace covers arbitrary shapes for the registered-provider case.
- *P2 alias precedence across injection* — `inject_settings_env_vars()` records what it injected; `resolve_provider_api_key` resolves genuinely exported vars (canonical-first) before settings-injected ones, so a live shell `GOOGLE_API_KEY` beats a stale settings `GEMINI_API_KEY`. Regression tests included.
- *P2 batch provider block* — batch item error dicts now carry `provider_message` end to end. Investigation found the motivating case doesn't even reach the exception path: `LLMNode` converts deterministic `LLMCallError`s to error-dict *results*, so `batch_executor` reads the value from the result's diagnostic context as well as from raised exceptions. Rendered as an indented `provider:` line (200-char cap) and preserved through JSON compaction. The suggestion wording is now position-independent ("shown with this error").
- *P2 effective model* — provider detection moved to `kwargs["model"]` after the `model_options` merge; tests cover registered→registered override (right key) and registered→unregistered (no key passed).

Also in this commit (separate concern, flagged in the description): default-model refresh — `gemini/gemini-3.5-flash-lite`, `anthropic/claude-sonnet-5` (incl. the hardcoded fallback, now the single source for the cache-analysis "pflow's default" annotation), `openai/gpt-5.6-luna`. All three verified present in LiteLLM's upstream catalog with the expected `litellm_provider`.

Field validation worth recording: the incident that motivated this PR resolved via `provider_message` — Google's raw text was "This model models/gemini-2.5-flash is no longer available to new users" (model retired for new accounts; the models list endpoint still lists it). Neither a name problem nor a key problem — exactly why the 404 suggestions now defer to the provider's own text.

Gates: `make check` clean, `make test` 9142 passed, `make test-e2e` 44 passed.

https://claude.ai/code/session_016ToQhpLMjSofEwaQcRdwVG

PR #614, opened 2026-08-23
fix: preserve MCP tools across partial sync failures (#612)

### LINKED ISSUE #612: MCP auto-sync deletes working tools and treats failed discovery as complete
## What happens

Automatic MCP discovery destructively clears the entire MCP registry before contacting any server. A transient discovery exception therefore deletes that server's last known-good tools. The run then records the configuration as synchronized even when discovery failed, so later ordinary runs skip discovery and the loss becomes sticky.

This is verified on current `main` at `81d2d56a` (2026-08-23), after the separate core-refresh fix in #611/#613.

## Verified mechanism

In `src/pflow/cli/mcp_sync.py`:

1. `_clean_old_mcp_entries()` deletes and persists every `mcp-*` entry before discovery (`:66-86`).
2. `_discover_and_register_servers()` registers a server only when `tools` is truthy (`:107-128`). Exceptions are recorded as failures, while a successful empty list is silently skipped.
3. `_auto_discover_mcp_servers()` calls the global cleanup before any server (`:170-180`).
4. It unconditionally writes `mcp_last_sync_time` and `mcp_servers_hash` after the loop (`:182-184`), even when `failed_servers` is non-empty.
5. `_check_mcp_sync_needed()` consequently returns false on the next run when config mtime/hash are unchanged (`:48-63`).

`MCPRegistrar.register_tools()` also adds/updates supplied tools without removing obsolete tools for that server (`src/pflow/mcp/registrar.py:49-94`). Removing the global wipe therefore requires successful discovery to replace one server's complete tool set, not merely add entries.

The current change detector is also unnecessarily coarse: any `mcp-servers.json` mtime change triggers every server, while its hash covers only the sorted server names. It cannot identify which server configuration changed.

## Executed reproduction

An isolated integration probe used the real `Registry`, real atomic persistence, and real `MCPRegistrar`, with only transport discovery outcomes injected.

### Partial failure

Starting entries: `good/old-tool`, `bad/old-tool`, and a removed server's old tool.

- `good` discovers `new-tool`.
- `bad` raises a temporary-outage exception.
- Persisted result: only `mcp-good-new-tool` remains.
- `bad/old-tool` was deleted.
- Completion timestamp and current server hash were stored.
- `_check_mcp_sync_needed()` returned `needs_retry = false`.

### Total failure

Starting entries: one old tool for each of two configured servers.

- Both discoveries raise temporary-outage exceptions.
- Persisted MCP entry count becomes zero.
- Completion timestamp and current server hash were stored.
- `_check_mcp_sync_needed()` returned `needs_retry = false`.

All 12 existing tests in `tests/test_cli/test_mcp_auto_discovery.py` pass on the buggy implementation; none asserts persisted preservation plus retry behavior after failure.

## Required behavior

Replace the global mtime/server-name-hash state with per-server configuration fingerprints stored in registry metadata.

- Fingerprint each raw persisted server configuration block independently using canonical JSON plus SHA-256. Hash stored `${VAR}` references, not resolved secret values.
- Formatting/key-order-only file changes do not trigger discovery. A command, args, env reference, URL, header, or other persisted value change triggers only that server.
- Added server: discover it; on success register its exact complete tool set and store its fingerprint.
- Changed server: discover only it; on success replace its exact complete tool set and advance only its fingerprint.
- Unchanged server: do not start it or rewrite its entries.
- Removed server: remove only entries canonically owned by that exact server and remove its fingerprint.
- Discovery exception: preserve that server's existing entries and prior fingerprint. It remains due on the next ordinary run.
- Successful empty discovery: replace that server's tool set with zero entries and advance its fingerprint.
- Successful servers advance independently of failed servers, so a retry contacts only servers still due.
- Manual `pflow mcp sync NAME` and `pflow mcp sync --all` are forced discovery operations but update fingerprints only for successful servers.
- Missing config file: preserve current registry state. Existing config with an empty `mcpServers` object: remove canonical MCP entries and store an empty fingerprint map.

Use `interface.mcp_metadata.server` as the sole ownership field for mutation, with exact equality. Do not infer ownership from node-name prefixes: `mcp-foo-` also matches a distinct `foo-bar` server. Current-created entries already carry canonical ownership. Backward compatibility for metadata-less legacy MCP entries is not required; full reconciliation may discard them rather than infer ownership.

Perform network discovery before loading the registry snapshot that will be committed. Recheck the relevant configuration after discovery; if it changed, publish nothing and retry later. Then load the latest unfiltered registry, apply removals and successful replacements in memory, preserve failures, and publish nodes plus the complete fingerprint map in one atomic registry write.

Interactive auto-sync output must name failed servers concisely. Non-interactive/structured output remains silent.

## Acceptance criteria

- Only added or changed servers are contacted; a formatting-only config rewrite contacts none and writes nothing.
- Partial failure: successful servers receive exact new tool sets and fingerprints; failed servers retain exact previous entries/fingerprints; the next run retries only failures.
- Total failure: every configured server retains its previous entries/fingerprints and remains retryable.
- Successful empty discovery removes obsolete tools and is not classified as failure.
- Removed-server cleanup and explicit removal use exact metadata ownership; changing/removing `foo` never changes `foo-bar`.
- An existing empty configuration cleans canonical MCP state; a missing config file preserves it.
- Manual single/all sync uses the same success/failure fingerprint rules as automatic sync.
- A configuration change during discovery causes no MCP reconciliation write or fingerprint advancement.
- Regression coverage uses real registry persistence for add/change/unchanged/remove, partial/total failure, retry, empty success, overlapping names, manual sync, and coherent nodes-plus-fingerprints persistence.

## Scope boundary

This issue removes the destructive global wipe, sticky failure state, stale same-server tools, and unnecessary all-server discovery. It does not add a general cross-process file lock, compare-and-swap protocol, or registry transaction system. Registry writes remain atomic complete-file replacements, but generic concurrent read-modify-write isolation is a separate concern.

## Environment

- pflow: current `main` (`81d2d56a`, version 0.15.1)
- Python: 3.13.4 for the verification probe
- OS: macOS; the defective control flow is platform-independent

spinje:
## Summary

Replaces destructive global MCP auto-sync with per-server raw-configuration fingerprints and one coherent post-discovery reconciliation. Successful servers advance independently, failed servers retain their last-known-good tools and fingerprint for retry, and unchanged servers are never started.

Closes #612

## Evaluation verdict

The issue remained reproducible on the branch base (`81d2d56a`): an executed real-Registry/real-MCPRegistrar probe left only the successful server's new tool after a partial discovery failure and reported no retry due. The issue text was revised before implementation to replace its original all-or-nothing global-marker proposal with per-server fingerprints, exact canonical ownership, missing-versus-empty config semantics, and one nodes-plus-fingerprints write.

The final shape is the simplest coherent form found in the existing architecture: discovery uses captured raw server blocks, the registrar owns one batch state machine, and `Registry.save(..., metadata_updates=...)` publishes nodes and fingerprint metadata in the same atomic replacement. It deliberately does not add a generic transaction abstraction.

## Changes

- Fingerprint each raw persisted server config with canonical JSON plus SHA-256; formatting, key order, and resolved secret-value changes are inert.
- Discover only added/changed servers and use the captured config block that was fingerprinted.
- Preserve failed servers, replace successful servers exactly (including successful empty results), and retry only servers still due.
- Remove absent servers and unsupported metadata-less MCP entries only during full reconciliation, using exact `interface.mcp_metadata.server` ownership.
- Share the same batch coordinator across workflow-start auto-sync, manual single sync, and manual `--all`; explicit removal deletes exact owned tools and its fingerprint coherently.
- Preserve registry state for a missing config and reconcile an existing explicit-empty config to no MCP entries/fingerprints.
- Extend `Registry.save()` narrowly so node and metadata updates share one atomic wrapper write.
- Update internal CLI/MCP/registry guidance and replace mock-only coverage with persisted state-machine regressions.

Backward compatibility for metadata-less legacy MCP entries is intentionally not provided, per the user ruling; full reconciliation discards them rather than guessing ownership from ambiguous node names. Generic cross-process registry read-modify-write locking/CAS remains explicitly outside #612.

## Review gate and dispositions

The external Claude fan-out was sandbox-blocked before sending repository content, so the documented direct-agent fallback ran six read-only lenses: silent failures, concurrency safety, feature interactions, impact completeness, test fidelity, and simplicity. A separate executing falsifier exercised the state table through real registry persistence, real stdio MCP processes, manual commands, and the actual `pflow run` entry path. Targeted re-reviews found no remaining in-scope findings.

Fixed findings:

- Forced `sync --all` selected targets before the coordinator snapshot. Verified by reading `src/pflow/cli/commands/mcp.py` and an executed regression that makes `manager.list_servers()` fail; forced-all now derives targets inside `MCPRegistrar.sync_servers()`.
- Manual `sync --all` could treat a missing config as explicit empty and delete MCP state. Reproduced with CliRunner + real Registry; fixed and pinned by `test_manual_sync_all_missing_config_preserves_registry`.
- Missing-file detection used a separate `exists()` check and load. Verified by concurrency review at `src/pflow/mcp/manager.py`; fixed with one sentinel-preserving raw snapshot API and pinned by `test_raw_server_snapshot_distinguishes_missing_from_explicit_empty`.
- Post-discovery schema conversion could abort the whole batch silently. Reproduced with two servers; replacement construction is now per-server transactional, and `test_conversion_failure_preserves_server_while_valid_peer_advances` proves the failed server is preserved while its peer commits.
- Raw SHA-256/environment-reference and manual single config-race contracts were unpinned. Fixed by independent known-digest, resolved-env/no-write, and manual abort/no-save tests.
- Cleanup policy, fingerprint decoding, and a mock rendering test had drift/duplication. Verified by the simplicity/test-fidelity lenses; consolidated and re-reviewed clean.

Skipped with reason:

- Three lenses reproduced the pre-existing flat node-key collision (`server=a-b, tool=c` versus `server=a, tool=b-c`). The binding audited plan explicitly excludes registry node-ID redesign; exact metadata ownership fixes overbroad removal/replacement for overlapping server names without claiming to solve that separate identity collision. The current change does not widen it.

## Verification

- Focused persisted-state suite: 124 passed across registry, config manager, registrar state machine, auto-sync, and manual MCP commands.
- `make check`: passed (Ruff, formatting, pre-commit, mypy, deptry, reusable-asset sync).
- `make test`: 9,173 passed.
- `make test-all-local`: 9,217 passed, 2 skipped.
- Real surface: with isolated HOME, `pflow run examples/simple-workflow.pflow.md --no-trace` started a real local stdio `pflow mcp serve`, persisted 13 canonically owned tools plus its fingerprint, and completed the workflow. A second unchanged run completed without changing the registry mtime.
- Executing falsifier: all tested promises held, including partial/total failure retries, empty success, config-change abort, late registry load, manual single/all sync, exact non-colliding overlapping-name removal, missing/explicit-empty behavior, and non-interactive output silence.

Platform-sensitive registry/config atomic replacement remains covered by the existing blocking Windows CI gate; local verification ran on macOS.

### DIFF
--- a/src/pflow/cli/CLAUDE.md
+++ b/src/pflow/cli/CLAUDE.md
@@ -211,7 +211,7 @@ Full list readable in `_initialize_context` and `_setup_workflow_execution` in `
 
 ## MCP Auto-Discovery (mcp_sync.py)
 
-Runs at startup on every `pflow` invocation. Smart skip: checks MCP config file mtime + SHA-256 hash of server list against stored metadata. Only re-syncs when config actually changed.
+Runs before workflow execution, not before other top-level commands. Each raw persisted server block has an independent SHA-256 fingerprint in registry metadata, so only added or changed servers start. Successful servers replace their exact owned tool set and advance independently; failures retain their tools/fingerprint for retry. A missing config is a no-op, while an explicit empty config reconciles MCP state to empty.
 
 Note: has its own copy of `_get_output_controller` (duplicated from `commands/run.py` to avoid circular imports).
 
@@ -264,7 +264,7 @@ See `core/CLAUDE.md` (shell_integration section) for FIFO detection, StdinData m
 | `error_output.py` | `test_unified_error_output.py`, `test_enhanced_error_output.py` |
 | `workflow_output.py` | `test_shell_stderr_warnings.py`, `test_direct_execution_helpers.py`, `test_workflow_output_source_simple.py` |
 | `workflow_resolution.py` | `test_workflow_resolution.py` |
-| `mcp_sync.py` | `test_mcp_auto_discovery.py` |
+| `mcp_sync.py` | `test_mcp_auto_discovery.py`, `test_mcp_auto_sync_state.py` |
 | `param_parsing.py` | `test_rerun_display.py`, `test_direct_execution_helpers.py` |
 | `commands/*` | See `commands/CLAUDE.md` |
 
--- a/src/pflow/cli/commands/CLAUDE.md
+++ b/src/pflow/cli/commands/CLAUDE.md
@@ -40,10 +40,10 @@ Subcommands: `list`, `find`, `describe`, `servers`, `add`, `sync`, `remove`, `se
 - `mcp find "description"` — LLM-powered MCP tool search via `find_components()`.
 - `mcp describe <tool>` — detailed tool info with parameters, outputs, and `.pflow.md` usage snippet.
 - `mcp servers` — lists configured **servers** (transport type, command/URL, timestamps). This was the old `mcp list`.
-- `mcp add|remove|sync` — server lifecycle management (unchanged).
+- `mcp add|remove|sync` — server lifecycle management. Manual sync is forced discovery but uses the same coherent per-server fingerprint/replacement path as workflow-start auto-sync.
 - `mcp serve` — launches pflow as an MCP server (stdio transport). Fundamentally different from the management commands.
 
-**Smart auto-discovery**: Runs at pflow startup on every command (via `mcp_sync.py`, not this file). Only syncs when config modified or servers changed.
+**Smart auto-discovery**: Runs only before workflow execution (via `mcp_sync.py`, not this file). Raw per-server config fingerprints prevent unchanged servers from starting; failures remain due without rolling back successful peers.
 
 **MCP tool normalization** (`normalize_node_id` in `registry/node_id.py`, 3-tier matching for `describe`/`probe`):
 1. Exact match: `mcp-slack-composio-SLACK_SEND_MESSAGE`
--- a/src/pflow/cli/commands/mcp.py
+++ b/src/pflow/cli/commands/mcp.py
@@ -4,11 +4,9 @@ Registered as a subgroup of the PflowCLI group in main.py via cli.add_command(mc
 Click handles subcommand routing natively.
 """
 
-import hashlib
 import json
 import logging
 import sys
-import time
 from typing import ClassVar
 
 import click
@@ -507,7 +505,7 @@ def remove(name: str, force: bool) -> None:
             sys.exit(1)
 
         # Count registered tools
-        registered_tools = registrar.list_registered_tools(name)
+        registered_tools = registrar.list_registered_tools(name, include_filtered=True)
 
         # Confirm removal
         if not force:
@@ -518,9 +516,9 @@ def remove(name: str, force: bool) -> None:
                 click.echo("Cancelled.")
                 return
 
-        # Remove tools from registry
-        if registered_tools:
-            removed = registrar.remove_server_tools(name)
+        # Remove exact owned tools and the server fingerprint before config removal.
+        removed = registrar.remove_server_tools(name)
+        if removed:
             click.echo(f"  Removed {removed} tools from registry")
 
         # Remove server configuration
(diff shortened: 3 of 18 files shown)


spinje:
Automatic-review dispositions on follow-up 0a2e4664:

- Fixed: manual single-server sync now handles the configuration disappearing before the coordinator snapshot and exits with an actionable error instead of indexing an empty result.
- Fixed: discovery now bounds the complete async MCP handshake/list operation with the existing per-server timeout setting and 30-second default. Executed proof used a real hung stdio child: configured 0.2s timeout, returned the timeout diagnostic in 2.209s including MCP subprocess cleanup, and the child PID was gone afterward.
- Fixed: removed the dead `removed_tools` and `registry_updated` result fields and their plumbing.
- Fixed suggestions: explicit-empty manual all-sync reports no configured servers after reconciliation; removed the unused manager parameter; consolidated fingerprint loading; documented replacement-before-removal; made conversion failures server/stage-specific; replaced the leaked `Any` result annotation; and made the auto-sync total count registered rather than merely discovered tools.
- Kept intentionally: the independent literal SHA-256 digest assertion. This guards the issue contract that fingerprints use canonical JSON SHA-256 rather than only comparing the helper to itself.
- Previously disposed: the flat MCP node-ID collision is a pre-existing registry identity limitation and the audited #612 scope explicitly excludes a node-ID redesign; this change uses exact canonical ownership and does not widen that limitation.

Verification on 0a2e4664: 50 focused tests passed; `make check` passed; `make test` passed 9,176 tests; real hung-stdio timeout probe passed.

spinje:
Windows CI disposition on follow-up 47c87a2f:

- Root cause (executed CI evidence): the new 30-second whole-discovery bound expired during the real Windows cold `npx @modelcontextprotocol/server-everything` startup. The immediately preceding passing run on 4543e2f took about 36.6 seconds from sync start to completion, proving the threshold—not the MCP path—caused the regression.
- Fixed: the bounded default is now 60 seconds, still within the reviewer-recommended 30–60 second range; explicit per-server `timeout` values continue to override it.
- Regression coverage: a focused test observes the 60-second cold-start default while the cancellation-based 0.01-second hung-server test continues to prove custom bounds.
- Local verification on 47c87a2f: 51 focused tests passed, `make check` passed, and `make test` passed all 9,177 tests.

The Windows MCP smoke check is the authoritative validation for the corrected platform seam and is rerunning on this head.

